Issue #15, opened 2024-11-30
Support breaking dependency cycles with provider/lazy


ZacSweers:
Probably need to add a way in `BindingStack` to denote a deferred type

PR #39, opened 2024-12-08
Introduce BindingEntry abstraction + move TypeMetadata into stack

ZacSweers:
Helps with #15 but doesn't solve it yet, that needs more work


ZacSweers:
Holding off on this for now. Feels like it complicates the impl and without any added features to show for it it doesn't feel like it's mergeable yet

PR #56, opened 2024-12-19
Implement Binds support

ZacSweers:
This implements support for functionality like Dagger `@Binds`. Providers must meet the following requirements to be candidates:

- A single extension receiver
- The receiver is a subtype of the exposed return type
- The property/function must be abstract
- The property/function must be private

Code gen will never call this function. Its implementation in code gen will throw an error.


ZacSweers:
One leftover thing — can we change the status of these to make them private/invisible references in FIR such that they're blocked from being called?

PR #60, opened 2024-12-20
Automatically generate assisted factories in FIR

### LINKED ISSUE #58: Consider generating default assisted factory interfaces
Can save a fair amount of boilerplate for these

ZacSweers:
Resolves #58

TODO:
- ~crashes on an owner issue when generating a function with assisted annotations~
- ~port default values over~


ZacSweers:
Finding that these don't seem to show up in the IDE, though my understanding is that they should. Put it behind an option for now and will continue with this later

Issue #51, opened 2024-12-18
Generate metadata onto generated factories


ZacSweers:
Currently if we generate a factory in IR, that is not visible in downstream compilations' IR phase because they have no metadata attached. We obviously need to be able to see these though, so let's attach this data somewhere

ZacSweers:
Folding into #86

PR #70, opened 2024-12-21
Update to Kotlin 2.1.20-Beta1

ZacSweers:


ZacSweers:
Going to save this one for when we do #51, as it'll be harder to test once this is updated

PR #110, opened 2025-01-11
Implement FIR graph generation

### LINKED ISSUE #32: Investigate generating implicit graph factories with FIR
The big question is how are instance/graph parameters propagated? I suspect they can't be

### LINKED ISSUE #86: Generate hidden FIR headers for generated classes
Necessary to work in multi-module.

Ideal case: we can expose in metadata directly in IR. This is preferable as it keeps the generated factories hidden.

Backup case: we generate header declarations for factories in FIR for metadata propagation.

Prior art: https://github.com/JetBrains/kotlin/blob/58750cdaf34f113be180f9839fd982250da69f16/plugins/plugin-sandbox/src/org/jetbrains/kotlin/plugin/sandbox/ir/MetadataExtensionEmitter.kt

Related to #51, but larger umbrella for adding tests too.

ZacSweers:
Effectively resolves #86, resolves #32


ZacSweers:
May need to switch to making the generation onto existing companion objects just generate regular functions without the supertype because some strange things happen with the introduction of supertype generation where it appears to break reading of callables from declarations added to. Downside is no private default parameter values but maybe that's fine as a caveat

PR #128, opened 2025-01-23
Use official compiler infrastructure

bnorm:
I've been really missing a lot of the functionality we have in the official Kotlin compiler test framework when testing out Lattice, so I put together this PR as a demonstration of how to set it up and create various tests. Essentially, you can have different kinds of categories of tests and then each test is just a `.kt` file. I've set up 3 categories here:

1. Diagnostic: create a `.kt` file somewhere under the `test/src/data/diagnostic` directory, rerun the `generateTests` Gradle task, and a new test will be generated for the file. When run, any diagnostics will be injected directly into the file as sort of a [golden value test](https://en.wikipedia.org/wiki/Characterization_test). Only the frontend of the compiler is executed, so these are extremely fast. I've ported 2 of your existing tests as examples.

2. Dump: create a `.kt` file somewhere under the `test/src/data/dump/fir` directory, rerun the `generateTests` Gradle... (you get the idea). When run, the entire FIR tree for the `.kt` file will be dumped to a separate file `.fir.txt`. This is extremely helpful to figure out what exactly is being generated by a plugin. Only the frontend is executed for these tests as well. And again, I've ported 2 of your existing tests as examples.

3. Box: When run, the file will be compiled and the `box()` function will be executed, expecting `"OK"` as the output. You have access to `kotlin.test.assert*` functions to validate behavior. These tests will run the entire compiler, so helpful to test IR changes as well. Once again, I've ported 2 of your tests.

In all of these tests, the biggest way you will modify their behavior is through [directives](https://github.com/JetBrains/kotlin/tree/master/compiler/test-infrastructure#directives). You can create your own - I've done so with `GENERATE_ASSISTED_FACTORIES` to enable that plugin option - or you can use many of the [existing directives](https://github.com/JetBrains/kotlin/tree/master/compiler/tests-common-new/tests/org/jetbrains/kotlin/test/directives). They are specified as line comments and can be simple boolean flags like `RENDER_DIAGNOSTICS_FULL_TEXT` or be more complex like disabling to rendering of certain diagnostics with `DIAGNOSTICS: -PROVIDES_OR_BINDS_SHOULD_BE_PRIVATE`. In each of the base tests, you can also configure default directives that you want with `defaultDirectives`. For example, the dump tests are basically diagnostics tests but with the `FIR_DUMP` diagnostic always enabled.

Also strongly recommend installing this plugin in IntelliJ; https://github.com/demiurg906/test-data-helper-plugin

Take a look, happy to answer any questions, and let me know if you want to move forward with this and I can port more tests.


ZacSweers:
I'm also not able to reproduce the CI failure locally 🤔

bnorm:
> I'm also not able to reproduce the CI failure locally 🤔

I have found the (action-junit-report)[https://github.com/mikepenz/action-junit-report] workflow action to be helpful to display test failures in CI.

> * It looks like the IDE plugin doesn't work with kotlin K2 mode :/

That's odd, never had an issue with the plugin in K2 mode before. Remember to select the "Test Data" view in the bottom left of the editor pane. Otherwise the file header with the buttons doesn't show up.

> * If we wanted to test a platform-specific behavior, how would we do that? For example - say we wanted to generate `@JvmStatic` on companion object creator functions in JVM compilations. Or say I want to test execution on target platforms.

Yes, this is all possible. The setup in the Kotlin repository is rather complex, so I can't really point you to anything right now. I'm working on simplify it for the piecemeal repository, both for the myself and as an example. Hoping to get JS, Wasm, and Native all working eventually.

And a single file can be run with all of them. You would create a new abstract base test and update the generator to point at the same files. And then if a test is backend specific or needs to ignore a backend, "TARGET_BACKEND" and "IGNORE_BACKEND" platform directives are available to control running of the test.

> * Say I wanted to test its interaction with a code gen tool like KSP. I assume that's probably not possible, but curious if it's come up!

It should be possible, but may require a little bit of extra configuration. Take a look at the `MetroExtensionRegistrarConfigurator` class to get an idea how you can manipulate the configuration of the compiler before test execution. You can depend on the KSP compiler-plugin directly in `compiler-tests`, and then you would need to configure it manually. Will it actually work? Not sure; we don't have any KSP tests in our test data, so I can't say for sure.

> * Is there a way to test incremental compilation?

We incremental compilation tests in Kotlin that use the old version of this test framework. I would assume they will be updated to use the new framework eventually, and any required functionality added. But for now I don't have any examples to show you.

> Super minor nit but can we put all of these under the `dev.zacsweers.metro.compiler` package instead of `dev.zacsweers.metro`?

You bet! I'll get everything moved.

PR #195, opened 2025-02-20
Fix wrapper type string formatting in FirContextualTypeKey

jisungbin:
Fixes an issue with extra "<" like `Provider<Lazy<<T>>` when isLazyWrappedInProvider type.


ZacSweers (approved):
Thanks!

PR #203, opened 2025-03-01
Implement exclusions and replacements

### LINKED ISSUE #136: Support `@Contributes*.replaces` arguments

ZacSweers:
Resolves #136


ZacSweers:
hmmm, this doesn't seem to work reliably. Finding a case in github.com/zacsweers/catchup where `replaces` arrays contain a mix of resolved and unresolved classes 🤔

ZacSweers:
The tricky part here is that we can't really move this to IR either because at that point we can't modify interfaces as far as I know. If we moved _all_ superinterface contributions to IR we could conditionally do it then, but that would be a pretty invasive change. I'm going to work with this for now and seek help from compiler folks in parallel.

Issue #210, opened 2025-03-06
Repeated ContributesBinding annotations result in duplicate declaration exception


japplin:
### Description

The follow snippet causes an exception during compilation:
```kotlin
interface Foo

@ContributesBinding(AppScope::class)
@ContributesBinding(LoggedInScope::class)
class FooBinding @Inject constructor() : Foo
```
Expected result would be to generate unique bindings for each scope.

### Self-contained Reproducer

Exception:
```
e: file:///Users/japplin/AndroidStudioProjects/MetroTesting/dagger-factory/src/main/java/com/metro/dagger/factory/FooBinding.kt:15:1 Platform declaration clash: The following declarations have the same JVM signature (getBindAsFoo(Lcom/metro/dagger/factory/FooBinding;)Lcom/metro/dagger/factory/Foo;):
    fun FooBinding.`<get-bindAsFoo>`(): Foo defined in com.metro.dagger.factory.FooBinding.`$$MetroContribution`
    fun FooBinding.`<get-bindAsFoo>`(): Foo defined in com.metro.dagger.factory.FooBinding.`$$MetroContribution`

```

### Metro version

1.0.0-SNAPSHOT

ZacSweers:
So in kotlin-inject-anvil, this is prohibited. Do you find this to be a common use case?

japplin:
Searching the register project at square we have roughly 20 instances of repeated ContributesBinding annotations, so I would say it's a somewhat common use case. The alternative of writing custom providers isn't a huge amount of boilerplate, so not supporting this could be workable for us. With that said though the annotation is currently marked as @Repeatable and this is a nice feature. I can look at contributing something for this issue as well if you're open to that.

edit. Actually looks like my initial search was not complete and we actually have close to 100 usages.

ZacSweers:
The repeatability is intended for contributing different bound types. That said, I'm open to this is there's an elegant way to do it. Currently contributing classes are all with the same name and 1:1 with contributing annotations. I think we could keep that and just have the asterisk that all contributions will be available to the union of scopes contributed to. I.e. still just one nested class MetroContribution class, but we can make it available to all scopes it has contributions for and ensure only one is created. Open to suggestions though if the need for for specific bindings to only specified scopes is needed though.

japplin:
Sorry, I accidentally marked this one as completed and it says I don't have permission to reopen

ZacSweers:
Done on main!

PR #212, opened 2025-03-06
Another test repeated ContributesBinding with different scopes and same bound type

japplin:
https://github.com/ZacSweers/metro/issues/210


compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/AggregationTest.kt
```diff
@@ -1195,6 +1195,45 @@ class AggregationTest : MetroCompilerTest() {
     }
   }
 
+  @Ignore("https://github.com/ZacSweers/metro/issues/210")
```
ZacSweers, line 1198:
```suggestion
```

ZacSweers (approved):

PR #222, opened 2025-03-11
Throws CliOptionProcessingException instead of IllegalStateException

jisungbin:
If the compiler argument parsing fails in a Kotlin CLI, a [CliOptionProcessingException is thrown](https://github.com/JetBrains/kotlin/blob/928416c9c5fe0287ed38c11ce3e93051c76eb598/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/plugins/PluginsOptionsParser.kt#L63).

```kotlin
open class CliOptionProcessingException(
  message: String, 
  cause: Throwable? = null,
) : RuntimeException(message, cause)
```


ZacSweers (approved):
Thanks!

PR #225, opened 2025-03-15
Fix a function typo in dagger interop

kevinguitar:
Spotted a typo when going through the dagger interop.

We plan to experiment Metro in our main BandLab Android app, we're currently using Dagger + Anvil KSP. We will see if we can adopt smoothly by keeping our custom KSP code gen, and provide some issues, contributions, and meaningful benchmarks. Thanks a lot for this framework 👍


ZacSweers (approved):
Thanks!

PR #228, opened 2025-03-19
Permit open classes to use @Inject annotation

japplin:
I was surprised to encounter an error that @Inject was not permitted for open classes, from looking at the code and local testing everything appears to be working and it was just this assertion preventing the use case. Any objections to allowing this use case?


ZacSweers (approved):
Never done this myself but open to it 👍

PR #233, opened 2025-03-20
Simplify testData file paths in TestDataPluginSettings

jisungbin:
By making the testData path in TestDataPluginSettings less specific, we can avoid having to update TestDataPluginSettings whenever new testData folders are added in the future.

This approach is also [used in Kotlin](https://github.com/JetBrains/kotlin/blob/a783f6594518b99a4478e2e5d6c7e10da72c9b34/.idea/kotlinTestDataPluginTestDataPaths.xml#L80), and everything works fine.


ZacSweers (approved):
heh, thanks

PR #239, opened 2025-03-21
Report additional information when duplicate bindings are detected

JoelWilcox:
Sometimes the contribution location (file + line number) are not available, which can make debugging and finding the duplicate difficult. We can still report the fully qualified class name of the contribution in at least some cases, or worst case we can print everything we know about the binding.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/BindingGraph.kt
```diff
@@ -43,29 +43,11 @@ internal class BindingGraph(private val metroContext: IrMetroContext) {
...
-              appendLine("├─ Binding ${i + 1}: $location")
-            }
-          }
-        }
+        appendLine("├─ Binding 1: ${existing.getContributionLocationOrDiagnosticInfo()}")
+        appendLine("├─ Binding 2: ${duplicate.getContributionLocationOrDiagnosticInfo()}")
         appendBindingStack(bindingStack)
       }
-      val location = binding.reportableLocation ?: bindingStack.graph.location()
+      val location = bindingStack.graph.location()
```
JoelWilcox, line 50: Not strictly necessary but I noticed that prioritizing the binding location here meant we were duplicating the info with the other binding info lines

ZacSweers (approved):
nice 👍

PR #242, opened 2025-03-23
Basic graph extensions

### LINKED ISSUE #165: Implement `@GraphExtension`
I'm a little conflicted on how to implement this at the moment. There are two ways I've been pondering

**Dagger style**

```kotlin
@DependencyGraph
interface AppGraph {
  val loggedInGraph: LoggedInGraph
}

@GraphExtension
interface LoggedInGraph
```

- Extensions would need to be declared (via exposing an accessor) in the parent graph, either directly or their creator.
- Can only create extensions from the parent
- Pro: More efficient code gen, nested classes can directly access parents' private provider functions, fields, etc
- Con: parent has to know directly about the extension as all its compilation is done there

**kotlin-inject style**

```kotlin
@DependencyGraph
interface AppGraph

@DependencyGraph
interface LoggedInGraph {
  @DependencyGraph.Factory
  interface Factory {
    fun create(@DependencyGraph parent: AppGraph)
  }
}
```

- Pro: No need for a separate annotation
- Pro: Parents don't need to know anything about extensions
- Pro: Extensions can be declared/compiled in later, downstream compilations
- Pro: Better isolation
- Possible pro: could extend multiple graphs?
- Con: parent still requires a real metro-generated graph. So while you get isolation in declaration, you don't really get isolation in usage
- Con: Less efficient code gen. In order to really support this, we would need to always expose accessors for every binding available in the parent graph.
  - kotlin-inject avoids this with its use of an internal public map for scoped providers and then lookups of provider functions. I think a better approach for metro would be exposing accessors + rely on tools like R8/proguard to remove unused accessors.
  - Could alleviate this by making extension opt-in, like `@DependencyGraph(extendable = true)`

ZacSweers:
Effectively resolves #165, but we didn't introduce a new annotation


ZacSweers:
**TODO**

- [x] Introduce `@Extends` annotation to opt these in
- [x] Introduce `@Includes` annotation for non-extension
- [x] Require one annotation on creator params. XOR `@Extends`, `@Provides`, and `@Includes`

PR #243, opened 2025-03-24
Make InstanceFactory a value class

ZacSweers:


runtime/src/jvmMain/kotlin/dev/zacsweers/metro/internal/InstanceFactory.jvm.kt
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2014 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metro.internal
+
+@JvmInline
```
bnorm, line 5: Why not add this annotation on the expect class so you don't even need the actual classes?

ZacSweers: Only available in the JVM artifact isn't it?

bnorm: It's available in common but only used during the JVM compilation.

ZacSweers: well that makes things easier!

ZacSweers: Continued in #297

PR #249, opened 2025-03-27
Convert all ProvidesErrorsTest tests to Kotlin infrastructure

bnorm:
Thus begins my quest to convert all of the tests. I'll go one test class at a time and try to avoid areas that are under active development for the time being.


bnorm:
Let me know if there's a test class you'd like to see converted next. Maybe one where you're not sure how some of the tests should be converted.

.gitattributes
```diff
@@ -1,4 +1,2 @@
 compiler-tests/**/*.java linguist-generated=true
-compiler-tests/**/*.fir.txt linguist-generated=true
-compiler-tests/**/*.fir.diag.txt linguist-generated=true
```
ZacSweers, line 3: Curious why you removed these, do you think they should be non-collapsed since they should be reviewed?

ZacSweers (approved):

.gitattributes
bnorm: Yes, I believe they should be reviewed. While they are initially generated by the test framework, if there is a change from an existing file, it is considered a test failure. Sort of like a golden value test.

The diag files are essentially the same as the assertions you had for the error/warning messages. And the fir files are essentially the same as checking for generated classes/functions in the class files.

PR #258, opened 2025-03-31
Add interop support for Anvil's ContributesMultibinding annotation

japplin:
This change adds interop support for dagger anvil's `@ContributeMultibinding` annotation. Annotated classes will behave the same as `@ContributesIntoSet`, UNLESS it is also annotated with `@MapKey`, in which case it will behave as `@ContributesIntoMap`.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ClassIds.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ClassIds.kt
@@ -15,6 +15,7 @@ public class ClassIds(
   customBindsAnnotations: Set<ClassId> = emptySet(),
   customContributesToAnnotations: Set<ClassId> = emptySet(),
   customContributesBindingAnnotations: Set<ClassId> = emptySet(),
+  internal val customContributesIntoSetAnnotations: Set<ClassId> = emptySet(),
   customElementsIntoSetAnnotations: Set<ClassId> = emptySet(),
   customGraphAnnotations: Set<ClassId> = emptySet(),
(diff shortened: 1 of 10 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ClassIds.kt
```diff
@@ -15,6 +15,7 @@ public class ClassIds(
   customBindsAnnotations: Set<ClassId> = emptySet(),
   customContributesToAnnotations: Set<ClassId> = emptySet(),
   customContributesBindingAnnotations: Set<ClassId> = emptySet(),
+  internal val customContributesIntoSetAnnotations: Set<ClassId> = emptySet(),
```
japplin, line 18: Not adding this annotation to `contributesIntoSetAnnotations` to facilitate additional logic to support `@MapKey`.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
```diff
@@ -65,7 +65,47 @@ internal object AggregationChecker : FirClassChecker(MppCheckerKind.Common) {
       if (classId in classIds.allContributesAnnotations) {
         val scope = annotation.resolvedScopeClassId() ?: continue
         val replaces = emptySet<ClassId>() // TODO implement
-
+        val checkIntoSet by lazy {
```
ZacSweers, line 68: nit: I have an `unsafeLazy` that acts like a simple memoize to skip the doubly-checked locking since this is never multithreaded

Issue #265, opened 2025-04-01
Support nullable bindings


ZacSweers:
Continuing from https://github.com/ZacSweers/metro/discussions/153, going to go with option 1 and treat `String` and `String?` as different type keys.

PR #275, opened 2025-04-04
Allow nullable types to be contributed to the graph

japplin:
Addresses issue #265


ZacSweers:
Some more work will be needed here heh, this just removes the FIR check but there's a bunch of places in the runtime that are currently null-averse (`T : Any` generics, etc) that will need updating. I've got a notepad written down somewhere I can share in a bit.

ZacSweers:
Continued in #333 and pulled some of your changes from here to that 👍

PR #277, opened 2025-04-04
Move Anvil tests to Kotlin test infrastructure

bnorm:
I saw #276 and thought I would show how to use external dependencies within the Kotlin test infrastructure. Here I'm loading the Anvil annotations if the new `WITH_ANVIL` directive is added. This adds the required jar files to the compile and runtime classpaths and adds annotations to the configuration. A similar pattern could be repeated for all other libraries you what to test interop with.

### DIFF
--- a/compiler-tests/build.gradle.kts
+++ b/compiler-tests/build.gradle.kts
@@ -5,7 +5,8 @@ plugins {
   java
 }
 
-val metroRuntimeClasspath: Configuration by configurations.creating
+val metroRuntimeClasspath: Configuration by configurations.creating { isTransitive = false }
+val anvilRuntimeClasspath: Configuration by configurations.creating { isTransitive = false }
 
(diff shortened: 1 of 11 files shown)


compiler-tests/build.gradle.kts
```diff
@@ -5,7 +5,8 @@ plugins {
   java
 }
 
-val metroRuntimeClasspath: Configuration by configurations.creating
+val metroRuntimeClasspath: Configuration by configurations.creating { isTransitive = false }
+val anvilRuntimeClasspath: Configuration by configurations.creating { isTransitive = false }
```
ZacSweers, line 9: For my own understanding, why the `isTransitive` addition here? Just to have dependencies explicitly added?

ZacSweers (approved):
Neat!

Issue #284, opened 2025-04-07
DependencyGraph doesn't pick up bindings from MemberInjector


kevinguitar:
### Description

I was playing around with the MemberInjector behavior and found this missing feature in metro.

### Self-contained Reproducer

Compilation failed with `Missing bindings for: [com.example.MyViewModel]`. With Dagger, `MyViewModel` factory will be included in the graph automatically because `MyScreen` member injector requires it.

```kotlin
@DependencyGraph
interface MyScreenGraph {
    fun inject(screen: MyScreen)

    @DependencyGraph.Factory
    interface Factory {
        fun create(@Provides screen: MyScreen): MyScreenGraph
    }
}

class MyScreen {
    @Inject lateinit var viewModel: MyViewModel

    init {
        createGraphFactory<MyScreenGraph.Factory>()
            .create(this)
            .inject(this)
    }
}

@Inject
class MyViewModel
```

### Metro version

0.1.1

PR #285, opened 2025-04-07
Failing test for graph to pick up member's dependency bindings

kevinguitar:
https://github.com/ZacSweers/metro/issues/284


ZacSweers:
ah didn't realize your branch was old, may need to spin this out to another PR

ZacSweers:
Fixed in #287

Issue #205, opened 2025-03-02
Contributes*.replaces is not reliable for references in the same compilation


ZacSweers:
See #203

ZacSweers:
We may need to just say that replaced types cannot be in the same compilation unit and call it a day

ZacSweers:
Documented

PR #288, opened 2025-04-07
Add ContributesBinding.rank interop support

JoelWilcox:
This specifically helps with migration for users coming from Dagger-Anvil since eliminating existing rank usages can get pretty tricky and time consuming depending on how many there are. There are a few limitation with handling this in FIR -- see ContributedInterfaceSupertypeGenerator#processRankBasedReplacements for a comment with all the details.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
@@ -6,20 +6,21 @@ import dev.zacsweers.metro.compiler.ClassIds
 import dev.zacsweers.metro.compiler.Symbols
 import dev.zacsweers.metro.compiler.fir.annotationsIn
 import dev.zacsweers.metro.compiler.fir.classIds
+import dev.zacsweers.metro.compiler.fir.rankValue
 import dev.zacsweers.metro.compiler.fir.resolvedAdditionalScopesClassIds
+import dev.zacsweers.metro.compiler.fir.resolvedBindingArgument
 import dev.zacsweers.metro.compiler.fir.resolvedExcludedClassIds
 import dev.zacsweers.metro.compiler.fir.resolvedReplacedClassIds
 import dev.zacsweers.metro.compiler.fir.resolvedScopeClassId
 import dev.zacsweers.metro.compiler.fir.scopeArgument
 import java.util.TreeMap
-import kotlin.sequences.forEach
+import kotlin.collections.plusAssign
 import org.jetbrains.kotlin.fir.FirAnnotationContainer
 import org.jetbrains.kotlin.fir.FirSession
 import org.jetbrains.kotlin.fir.analysis.checkers.getContainingClassSymbol
 import org.jetbrains.kotlin.fir.caches.FirCache
 import org.jetbrains.kotlin.fir.caches.firCachesFactory
 import org.jetbrains.kotlin.fir.declarations.FirClassLikeDeclaration
-import org.jetbrains.kotlin.fir.declarations.utils.classId
 import org.jetbrains.kotlin.fir.expressions.FirAnnotation
 import org.jetbrains.kotlin.fir.extensions.FirSupertypeGenerationExtension
 import org.jetbrains.kotlin.fir.extensions.predicate.LookupPredicate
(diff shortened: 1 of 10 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -215,6 +220,88 @@ internal class ContributedInterfaceSupertypeGenerator(
...
+   *    implicit type.
+   * 2. We can't check for qualifiers when comparing these bindings because those annotation calls
+   *    are not resolved at this point. E.g. a compiler critical annotation
+   *    like @ContributesBinding(..) will be resolved but @Named(..) will not. This means that
+   *    qualifiers are unsupported for rank interop support. Rank can only effectively be used for
+   *    binding types where all of them are unqualified or all use the same qualifier. Other
+   *    combinations will need to be migrated to instead use explicit replacements or exclusions.
+   *
+   *     @return The bindings which have been outranked and should not be included in the merged
+   *       graph
```
ZacSweers, line 260:
```suggestion
   * @return The bindings which have been outranked and should not be included in the merged graph.
```

ZacSweers (approved):
Nice!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: Oop, fixed 😁

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -215,6 +220,87 @@ internal class ContributedInterfaceSupertypeGenerator(
...
+   * 1. Ranked bindings and bindings that get outranked must both explicitly declare their binding
+   *    type in order for us to actually compare them, because supertypes are not resolvable here.
+   *    The user will end up getting a duplicate binding error if the outranked binding is using an
+   *    implicit type.
+   * 2. We can't check for qualifiers when comparing these bindings because those annotation calls
+   *    are not resolved at this point. E.g. a compiler critical annotation
+   *    like @ContributesBinding(..) will be resolved but @Named(..) will not. This means that
+   *    qualifiers are unsupported for rank interop support. Rank can only effectively be used for
+   *    binding types where all of them are unqualified or all use the same qualifier. Other
+   *    combinations will need to be migrated to instead use explicit replacements or exclusions.
```
bnorm, line 257: This may not be a true limitation. You can create use DeclarationPredicate to define meta-annotations.
https://github.com/JetBrains/kotlin/blob/52a3ec9184fa44b2c8ce981f279cd66686dbe73b/plugins/plugin-sandbox/src/org/jetbrains/kotlin/plugin/sandbox/fir/SomeAdditionalSupertypeGenerator.kt#L28-L33

This should cause anything annotated with the meta-annotation to also become a "compiler critical annotation". (Registering the predicates is what causes the annotations to become "critical", meta-predicates just automatically register anything they annotate)

ZacSweers: Iiiinteresting. Do you think that would support this case to? https://github.com/ZacSweers/metro/issues/205

bnorm: I may not fully understand what the problems with #205 where, but I don't think so. I don't think it will cause the class references to `ContributesTo`-annotated interfaces to become resolved. I suspect it only impacts the resolution of annotations that are annotated.

But you could certainly try it! I don't have direct knowledge about what to expect.

bnorm: You could hard-pivot and require all scopes to be annotation classes! 😅

JoelWilcox: Wow I didn't know there were mechanisms for meta annotations. Will give this a try once I have a chance today!

bnorm: [Meta-meta-meta-annotations too!](https://github.com/JetBrains/kotlin/blob/adc0a0eb422ad7cca2be8fe798b2d5383a3a27ce/plugins/plugin-sandbox/testData/diagnostics/supertypes/metaAnnotationOrder.kt#L34-L45)

PR #292, opened 2025-04-07
Include metro version in metadata

### LINKED ISSUE #274: Include a Metro version in MetroMetadata
This'll be helpful in the future if we ever process older metadata and want to gracefully degrade

ZacSweers:
Resolves #274


compiler/src/main/proto/dev/zacsweers/metro/compiler/proto/metro_metadata.proto
ZacSweers, line 1: Breaking change but this is 0.x days

PR #293, opened 2025-04-08
Support qualifiers and implicit boundTypes in dagger-anvil rank interop

JoelWilcox:
Follow-up from #288. Thanks @bnorm for catching and explaining how this could be handled!

edit: This PR also adds support for implicit boundTypes in the rank interop thanks to a good catch from @ZacSweers that it's possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ _2025-04-08_
 - **Change**: Change `InstanceFactory` to a value class.
 - **Change**: Make `providerOf` use `InstanceFactory` under the hood.
 
-Special thanks to [@JoelWilcox](https://github.com/JoelWilcox), [@bnorm](https://github.com/bnorm), [@japplin](https://github.com/japplin), [@kevinguitar](https://github.com/kevinguitar), and [@erawhctim](https://github.com/erawhctim) for contributing to this release! 
+Special thanks to [@JoelWilcox](https://github.com/JoelWilcox), [@bnorm](https://github.com/bnorm), [@japplin](https://github.com/japplin), [@kevinguitar](https://github.com/kevinguitar), and [@erawhctim](https://github.com/erawhctim) for contributing to this release!
 
 0.1.1
 -----
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/ExtensionPredicates.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/ExtensionPredicates.kt
@@ -1,0 +1,23 @@
+// Copyright (C) 2025 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metro.compiler.fir
+
+import dev.zacsweers.metro.compiler.ClassIds
+import dev.zacsweers.metro.compiler.asFqNames
+import org.jetbrains.kotlin.fir.extensions.predicate.DeclarationPredicate
+import org.jetbrains.kotlin.fir.extensions.predicate.LookupPredicate
+
+public class ExtensionPredicates(private val classIds: ClassIds) {
+
+  // Lets us register and resolve any annotations that are qualifiers
+  internal val qualifiersPredicate =
+    DeclarationPredicate.create {
+      metaAnnotated(classIds.qualifierAnnotations.asFqNames(), includeItself = false)
+    }
+
+  internal val dependencyGraphPredicate =
+    LookupPredicate.create { annotated(classIds.dependencyGraphAnnotations.asFqNames()) }
+
+  internal val contributingTypesPredicate =
+    LookupPredicate.create { annotated(classIds.allContributesAnnotations.asFqNames()) }
+}
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/ExtensionPredicates.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/ExtensionPredicates.kt
@@ -7,7 +7,7 @@ import dev.zacsweers.metro.compiler.asFqNames
 import org.jetbrains.kotlin.fir.extensions.predicate.DeclarationPredicate
 import org.jetbrains.kotlin.fir.extensions.predicate.LookupPredicate
 
-public class ExtensionPredicates(private val classIds: ClassIds) {
+internal class ExtensionPredicates(private val classIds: ClassIds) {
 
   // Lets us register and resolve any annotations that are qualifiers
   internal val qualifiersPredicate =
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirExtensionRegistrar.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirExtensionRegistrar.kt
@@ -23,15 +23,16 @@ import org.jetbrains.kotlin.fir.extensions.FirSupertypeGenerationExtension
 
 public class MetroFirExtensionRegistrar(
   private val classIds: ClassIds,
+  private val predicates: ExtensionPredicates,
   private val options: MetroOptions,
 ) : FirExtensionRegistrar() {
   override fun ExtensionRegistrarContext.configurePlugin() {
-    +MetroFirBuiltIns.getFactory(classIds, options)
+    +MetroFirBuiltIns.getFactory(classIds, predicates, options)
     +::MetroFirCheckers
     +supertypeGenerator("Supertypes - graph factory", ::GraphFactoryFirSupertypeGenerator, false)
     +supertypeGenerator(
       "Supertypes - contributed interfaces",
-      ContributedInterfaceSupertypeGenerator.Factory(classIds)::create,
+      ::ContributedInterfaceSupertypeGenerator,
       false,
     )
     +supertypeGenerator(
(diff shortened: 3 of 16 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -67,9 +69,13 @@ internal class ContributedInterfaceSupertypeGenerator(
...
   private val contributingTypesPredicate =
-    LookupPredicate.create {
-      annotated(classIds.allContributesAnnotations.map { it.asSingleFqName() })
+    LookupPredicate.create { annotated(classIds.allContributesAnnotations.asFqNames()) }
+
+  private val qualifiersPredicate by unsafeLazy {
+    DeclarationPredicate.create {
+      metaAnnotated(classIds.qualifierAnnotations.asFqNames(), includeItself = false)
     }
+  }
```
bnorm, line 78: Kind of nit: predicates are cheap. And don't even need to be part of the extension itself. So (1),don't need `unsafeLazy` and (2), consider putting all the predicates in a shared class like the `ClassId`s are now.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -138,6 +144,13 @@ internal class ContributedInterfaceSupertypeGenerator(
     return graphAnnotation.scopeArgument() != null
   }
 
+  override fun FirDeclarationPredicateRegistrar.registerPredicates() {
+    // Currently we only need to resolve qualifiers for dagger-anvil's 'rank' interop
```
bnorm, line 148: All the predicates should be registered if they are used for lookups. The other two predicates are only working because another extension registers a similar predicate using the same annotations. Registration is cheap as well, so every extension should really register the extensions it uses, to avoid breakage due to another extension changing what it registers.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: This is the best type of nit 😁, I wasn't sure how expensive the predicates were before so it's helpful context going forward. Updating to move all the predicates used here to a unified file, and can put up a followup PR to migrate the other extensions to use it as well

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: Thanks! I was curious how that was working since they weren't already being registered here

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ _2025-04-08_
 - **Change**: Change `InstanceFactory` to a value class.
 - **Change**: Make `providerOf` use `InstanceFactory` under the hood.
 
-Special thanks to [@JoelWilcox](https://github.com/JoelWilcox), [@bnorm](https://github.com/bnorm), [@japplin](https://github.com/japplin), [@kevinguitar](https://github.com/kevinguitar), and [@erawhctim](https://github.com/erawhctim) for contributing to this release! 
+Special thanks to [@JoelWilcox](https://github.com/JoelWilcox), [@bnorm](https://github.com/bnorm), [@japplin](https://github.com/japplin), [@kevinguitar](https://github.com/kevinguitar), and [@erawhctim](https://github.com/erawhctim) for contributing to this release!
```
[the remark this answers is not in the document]
ZacSweers: 😅

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/ExtensionPredicates.kt
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2025 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metro.compiler.fir
+
+import dev.zacsweers.metro.compiler.ClassIds
+import dev.zacsweers.metro.compiler.asFqNames
+import org.jetbrains.kotlin.fir.extensions.predicate.DeclarationPredicate
+import org.jetbrains.kotlin.fir.extensions.predicate.LookupPredicate
+
+public class ExtensionPredicates(private val classIds: ClassIds) {
```
ZacSweers, line 10: nice. Let's keep these internal though!

```suggestion
internal class ExtensionPredicates(private val classIds: ClassIds) {
```

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirExtensionRegistrar.kt
```diff
@@ -23,15 +23,16 @@ import org.jetbrains.kotlin.fir.extensions.FirSupertypeGenerationExtension
...
 ) : FirExtensionRegistrar() {
   override fun ExtensionRegistrarContext.configurePlugin() {
-    +MetroFirBuiltIns.getFactory(classIds, options)
+    +MetroFirBuiltIns.getFactory(classIds, predicates, options)
     +::MetroFirCheckers
     +supertypeGenerator("Supertypes - graph factory", ::GraphFactoryFirSupertypeGenerator, false)
     +supertypeGenerator(
       "Supertypes - contributed interfaces",
-      ContributedInterfaceSupertypeGenerator.Factory(classIds)::create,
+      ::ContributedInterfaceSupertypeGenerator,
```
ZacSweers, line 35: Nice

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -244,17 +240,10 @@ internal class ContributedInterfaceSupertypeGenerator(
...
-   * 2. We can't check for qualifiers when comparing these bindings because those annotation calls
-   *    are not resolved at this point. E.g. a compiler critical annotation
-   *    like @ContributesBinding(..) will be resolved but @Named(..) will not. This means that
-   *    qualifiers are unsupported for rank interop support. Rank can only effectively be used for
-   *    binding types where all of them are unqualified or all use the same qualifier. Other
-   *    combinations will need to be migrated to instead use explicit replacements or exclusions.
+   * The main limitation we have is that supertypes are not resolvable here. That means ranked
+   * bindings and bindings that get outranked must both explicitly declare their binding type for us
+   * to actually compare them. The user will end up getting a duplicate binding error if the
+   * outranked binding is using an implicit type.
```
ZacSweers, line 246: Can you unpack this one a bit more for me here? Or maybe add an example snippet to the kdoc here?

ZacSweers (commented):
Couple nits!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: Took a stab at making this clearer, lmk if it's helpful or could use more clarification

ZacSweers: lgtm!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -311,6 +294,24 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
...
+        (this as FirClassSymbol<*>).fir.superTypeRefs.map {
+          typeResolver.resolveUserType(it as FirUserTypeRef).coneType
+        }
+      } else {
+        (this as FirClassSymbol<*>).resolvedSuperTypes
+      }
+      .singleOrError {
+        val superTypeFqNames = map { it.classId?.asSingleFqName() }.joinToString()
+        "${classId.asSingleFqName()} has a ranked binding with no explicit bound type and $size supertypes ($superTypeFqNames). There must be exactly one supertype or an explicit bound type."
+      }
```
ZacSweers, line 312: Good with merging as-is but these `.fir` refs are probably internal APIs we should be avoiding iirc. Usually there's extension helpers to avoid it. Let me poke around a bit

ZacSweers: tried a bit, I see the difficulty now, giving up until a later time

ZacSweers (approved):

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: Aye this section gets a bit tricky. Not sure of alternatives for `fir.superTypeRefs`, but we could probably replace the direct `fir.resolveState` reference by switching to something like a try-catch that first tries to access `resolvedSuperTypes` and falls back to the unresolved path

Issue #98, opened 2025-01-05
Explore IC optimizations for `@Contributes*Binding`


ZacSweers:
With first-party binding contribution support, I think we might be able to scrounge out some IC enhancements. The goal would be to allow implementation classes to remain `internal` or even private, while having bindings only expose their factories. 

## Design

Based on this example

```kotlin
@ContributesBinding<Cache>(AppScope::class)
@Inject
internal class CacheImpl(val fs: FileSystem): Cache
```

Impl class can be non-public

Generated factories would need to hide their references to the impl class.

```diff
class $$LatticeFactory(
  private val fs: Provider<FileSystem>
-) : Factory<CacheImpl> {
+) : Factory<Cache> {
-  override fun invoke(): CacheImpl {
+  override fun invoke(): Cache {
    return newInstance(fs())
  }

  companion object {
    // ...
    
-    fun newInstance(fs: FileSystem): CacheImpl {
+    fun newInstance(fs: FileSystem): Cache {
      return CacheImpl(fs)
    }
  }
}
```

However, this breaks down pretty quick with multiple bindings. So maybe instead we create some sort of companion accessor factory in the same file (which could access these factories still as "private to this"). They would expose all the same dependencies but just delegate to the real factory under the hood and be untyped (i.e. `Factory<*>`). Then generated binding hints could point at these factories. It's probably easiest to just keep them untyped and cast so that they can be reused.

This is as far as I've thought about this so far.

minirogue:
This is the first hurdle I ran into trying to use Metro. I prefer to keep all of my implementations `internal`, which is supported by Dagger (I'm unsure for other DI frameworks). Can you expand on what the issue is with multiple bindings?

As a short-term improvement, it might be helpful to provide a more direct error message in these cases. Right now this is what I got during a `compileKotlin` task:

> e: file:///path/to/AppDependencyGraph.kt:7:1 'public' sub-interface exposes its 'internal' supertype containing declaration 'InternalImplementationClassName'.

Which was unclear to me, since `AppDependencyGraph` is not overtly a sub-interface (or super-interface) and I expected the framework to hide the implementation itself from the scope it's contributing to. We should be able to throw an error on the `internal class` that is attempting to contribute its binding at the point that the factory is getting generated, right?

ZacSweers:
> which is supported by Dagger

[...]

PR #294, opened 2025-04-08
Add visibility checks

Minirogue:
Rough draft as of now. Just looking for feedback on comparing the visibilities of contributed classes vs. the scope they're being contributed to. If a scope has access to a class that has stricter visibility, then that scope could leak the class and cause a compilation error outside of our control.

See #98 and #295 for context.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
```diff
@@ -3,6 +3,8 @@
 package dev.zacsweers.metro.compiler.fir.checkers
 
 import dev.drewhamilton.poko.Poko
+import dev.zacsweers.metro.compiler.asName
+import dev.zacsweers.metro.compiler.fir.*
```
ZacSweers, line 7: nit: please match the code style of the project

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
```diff
@@ -169,10 +174,32 @@ internal object AggregationChecker : FirClassChecker(MppCheckerKind.Common) {
...
+        declaration: FirClass,
+        annotation: FirAnnotation,
+        session: FirSession,
+        context: CheckerContext,
+        reporter: DiagnosticReporter
+    ): Boolean {
+        val scopeVisibility =
+            annotation.getKClassArgument(name = "scope".asName(), session)?.toClassSymbol(session)?.visibility
+        val declarationVisibility = declaration.visibility
+        val visibilityComparison = scopeVisibility?.let { declarationVisibility.compareTo(it) }
```
ZacSweers, line 192: Just double checking - are you sure these are (intentionally) ordered by visibility?

ZacSweers (commented):
Overall a fan of this, couple nits

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
Minirogue: From `org.jetbrains.kotlin.descriptors.Visibilities`, this is the value map used to make the comparison:
Screenshot 2025-04-09 at 8 07 31 PM

ZacSweers: But you're not using that class here correcty? Shouldn't this be

```suggestion
        val visibilityComparison = scopeVisibility?.let { Visibilities.compare(declarationVisibility, it) }
```

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
ZacSweers: I know, spotless doesn't check this unfortunately. Just asking that you match it anyway :)

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AggregationChecker.kt
Minirogue: Not using it directly, but `Visibility.compareTo()` calls `Visibilities.compareLocal()` which uses the map in the screenshot to make the comparison.
Screenshot 2025-04-09 at 9 09 42 PM
Screenshot 2025-04-09 at 9 09 51 PM

Minirogue: Oh weird, I just noticed `Visibilities.compare()`, which calls `Visibility.compareTo()` (which calls `Visibilities.compareLocal()`) and then reverses the comparison on a `null`... So I guess maybe that's better? I'm not really sure when `Visibilities.compare()` would actually use `oppositeResult`, though.
Screenshot 2025-04-09 at 9 17 17 PM

ZacSweers: I'm fine with either as long as it's correct!

ZacSweers:
@Minirogue are you still planning to finish this?

ZacSweers:
I'll just finish this up myself 👍

ZacSweers:
hmm, actually upon further reading I think this check tries to solve the wrong problem. The issue with visibility is catching if an `internal` type is is contributed to a `public` graph. The scope key itself is sort of irrelevant to that situation. Let me know if I'm misunderstanding your goal in this PR though?

minirogue:
> @Minirogue are you still planning to finish this?

Sorry, it fell to my backburner. I've got style fixes, but I wanted to at least reinforce the testing before opening it up.

> hmm, actually upon further reading I think this check tries to solve the wrong problem. The issue with visibility is catching if an `internal` type is is contributed to a `public` graph. The scope key itself is sort of irrelevant to that situation. Let me know if I'm misunderstanding your goal in this PR though?

My original plan and reasoning was this (colored entirely by my approach to DI, so I may not be accounting for some usecases/patterns):
- Bindings get contributed to scopes and dependency graphs access those bindings through the scope.
- So, anything that can see a scope should be able to see all of its bindings and
- anything that can see a dependency graph should be able to see any scopes it uses.

What I have so far is only half of my proposed solution (the second bullet point), but at least fixes the issue when using `AppScope`, since it is public. I wasn't able to nail down the dependency graph vs. scope visibilities (third bullet) as quickly, so either that still needs to be added to this PR still or done separately.

ZacSweers:
I don't think this threads the right needle for what we want, versus just a checker that ensures that no merged type's contributed supertypes have narrower visibility than the type that's merged it. Going by the scope keys alone isn't a good heuristic as they're often publicly defined in util modules

ZacSweers:
Closing in favor of #347, which checks supertypes directly

PR #297, opened 2025-04-08
Make InstanceFactory a value class

ZacSweers:


ZacSweers:
Questioning reality a bit

image

PR #308, opened 2025-04-10
Fix repeatable @ContributesTo annotations being missed when aggregating contributions from downstream modules

JoelWilcox:


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -118,6 +119,20 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
...
 
+  private fun FirRegularClassSymbol.contributesAnnotations(): Sequence<FirAnnotation> {
+    return annotations.annotationsIn(session, session.classIds.allContributesAnnotations).ifEmpty {
+      // When there are multiple, they end up getting bundled into a 'Container' type that we need
+      // to unwrap. This behavior appears to only occur for contributions from other
+      // modules/compilations.
+      ((annotations.firstOrNull()?.argumentMapping?.mapping?.values?.firstOrNull()
+            as? FirArrayLiteral)
+          ?.argumentList
+          ?.arguments as? List<FirAnnotation>)
```
JoelWilcox, line 130: I feel like there's probably a better way to do this but I didn't see anything else that worked. @bnorm do you know of any helpers that might be useful here?

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/AggregationTest.kt
```diff
@@ -898,6 +890,110 @@ class AggregationTest : MetroCompilerTest() {
...
+    }
+  }
+
+  private fun Class<*>.assertHasSupertype(superTypeFqName: String) {
+    assertThat(allSupertypes().map { it.name })
+      .containsExactly("$superTypeFqName$$\$MetroContribution", superTypeFqName)
+  }
+
+  @Test
+  fun `ContributesTo can be repeated to contribute to multiple scopes in a merging module`() {
```
JoelWilcox, line 959: Note: this use-case already worked but I noticed it wasn't explicitly covered by a test yet

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
ZacSweers: I did add an `.expectAs()` helper to at least save some of the parens 🥲

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/AggregationTest.kt
```diff
@@ -898,6 +890,110 @@ class AggregationTest : MetroCompilerTest() {
...
+      assertThat(altGraph.callProperty<String>("altVal")).isEqualTo("Hello, world!")
+
+      val thirdGraphClass = classLoader.loadClass("test.ThirdGraph")
+      val thirdGraph = thirdGraphClass.generatedMetroGraphClass().createGraphWithNoArgs()
+      thirdGraphClass.assertHasSupertype("test.ContributedInterface")
+      assertThat(thirdGraph.callProperty<String>("thirdVal")).isEqualTo("Hello, world!")
+    }
+  }
+
+  private fun Class<*>.assertHasSupertype(superTypeFqName: String) {
```
ZacSweers, line 953: nit: let's call this `assertHasContributedSupertype` to make it more obvious that this is going to look for the `$$MetroContribution` type

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -118,6 +119,20 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
       }
     }
 
+  private fun FirRegularClassSymbol.contributesAnnotations(): Sequence<FirAnnotation> {
+    return annotations.annotationsIn(session, session.classIds.allContributesAnnotations).ifEmpty {
+      // When there are multiple, they end up getting bundled into a 'Container' type that we need
+      // to unwrap. This behavior appears to only occur for contributions from other
+      // modules/compilations.
```
ZacSweers, line 126: Should we gait this check to just if this class symbol is external?

JoelWilcox: That sounds fair to me! Didn't realize that property was available 👀

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: The chain simultaneously looks better and crazier now 😂

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: Hmm actually it looks like `isExternal` isn't compatible with this use-case -- the prop is false for the new test case when the downstream module contributions get evaluated

ZacSweers: Ah no you don't want that. That refers to the `external` keyword (opposite of `internal` but much rarer use. Analogous to `native` in Java). For here I think you can just check the key of the symbol and see if it's an external declaration/stub?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
bnorm: There are a couple in the compiler I found; `unwrapAndFlattenArgument` and `unwrapVarargValue` for example. But I think it might be better for compatibility with future Kotlin compiler versions if you write your own utility using those as inspiration.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -118,6 +121,30 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
...
+          ?.argumentMapping
+          ?.mapping
+          ?.values
+          ?.firstOrNull()
+          ?.expectAsOrNull<FirArrayLiteral>()
+          ?.argumentList
+          ?.arguments
+          ?.expectAsOrNull<List<FirAnnotation>>()
+          .orEmpty()
+          .annotationsIn(session, session.classIds.allContributesAnnotations)
```
ZacSweers, line 141: image

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -118,6 +121,30 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
       }
     }
 
+  private fun FirRegularClassSymbol.contributesAnnotations(): Sequence<FirAnnotation> {
+    return annotations.annotationsIn(session, session.classIds.allContributesAnnotations).ifEmpty {
+      if (origin == FirDeclarationOrigin.Library) {
```
ZacSweers, line 126: @bnorm question for later - is this the right way in FIR to check if it's from an external source?

ZacSweers (approved):

PR #321, opened 2025-04-15
Support repeatable @ContributesBinding annotations with different scopes

JoelWilcox:
- Addresses https://github.com/ZacSweers/metro/issues/210
- @ContributesBinding could already be repeated to contribute multiple types to a single scope. These changes expand the support to include bindings for different scopes, bringing behavior closer to Dagger-Anvil.
- A notable implementation change is that previously classes with contributions would have a 1:1 relationship where each class has a nested $$MetroContribution class created. Now, a class with contributions will have a nested class for each scope that it is contributed to. These nested classes will be sequentially numbered like `$$MetroContribution`, and `$$MetroContribution2`.
- Introduces `@MetroContribution` which is used to annotate each of the nested contribution classes and track the scope that they are associated with.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
@@ -6,7 +6,6 @@ import dev.zacsweers.metro.compiler.Symbols
 import dev.zacsweers.metro.compiler.asName
 import dev.zacsweers.metro.compiler.capitalizeUS
 import dev.zacsweers.metro.compiler.decapitalizeUS
-import dev.zacsweers.metro.compiler.expectAs
 import dev.zacsweers.metro.compiler.expectAsOrNull
 import dev.zacsweers.metro.compiler.mapToArray
 import java.util.Objects
@@ -64,7 +63,6 @@ import org.jetbrains.kotlin.fir.expressions.builder.buildAnnotationArgumentMappi
 import org.jetbrains.kotlin.fir.expressions.builder.buildAnnotationCall
 import org.jetbrains.kotlin.fir.expressions.builder.buildEnumEntryDeserializedAccessExpression
 import org.jetbrains.kotlin.fir.expressions.builder.buildLiteralExpression
-import org.jetbrains.kotlin.fir.expressions.toReference
 import org.jetbrains.kotlin.fir.expressions.unexpandedClassId
 import org.jetbrains.kotlin.fir.extensions.FirSupertypeGenerationExtension.TypeResolveService
 import org.jetbrains.kotlin.fir.extensions.buildUserTypeFromQualifierParts
@@ -857,16 +855,6 @@ internal fun FirAnnotation.resolvedScopeClassId(typeResolver: TypeResolveService
   // try to resolve within the enclosing scope
   return scopeArgument.resolvedClassId()
     ?: scopeArgument.resolvedClassArgumentTarget(typeResolver)?.classId
-}
-
-internal fun FirAnnotation.scopeName(session: FirSession): String? {
-  return resolvedScopeClassId()?.shortClassName?.identifier
-    ?: scopeArgument()
-      ?.argument
-      ?.toReference(session)
-      ?.expectAs<FirSimpleNamedReference>()
-      ?.name
-      ?.identifier
 }
 
 internal fun FirAnnotation.resolvedAdditionalScopesClassIds() =
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
@@ -6,6 +6,7 @@ import dev.zacsweers.metro.compiler.Symbols
 import dev.zacsweers.metro.compiler.asName
 import dev.zacsweers.metro.compiler.capitalizeUS
 import dev.zacsweers.metro.compiler.decapitalizeUS
+import dev.zacsweers.metro.compiler.expectAs
 import dev.zacsweers.metro.compiler.expectAsOrNull
 import dev.zacsweers.metro.compiler.mapToArray
 import java.util.Objects
@@ -63,6 +64,7 @@ import org.jetbrains.kotlin.fir.expressions.builder.buildAnnotationArgumentMappi
 import org.jetbrains.kotlin.fir.expressions.builder.buildAnnotationCall
 import org.jetbrains.kotlin.fir.expressions.builder.buildEnumEntryDeserializedAccessExpression
 import org.jetbrains.kotlin.fir.expressions.builder.buildLiteralExpression
+import org.jetbrains.kotlin.fir.expressions.toReference
 import org.jetbrains.kotlin.fir.expressions.unexpandedClassId
 import org.jetbrains.kotlin.fir.extensions.FirSupertypeGenerationExtension.TypeResolveService
 import org.jetbrains.kotlin.fir.extensions.buildUserTypeFromQualifierParts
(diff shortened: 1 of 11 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
```diff
@@ -857,6 +859,16 @@ internal fun FirAnnotation.resolvedScopeClassId(typeResolver: TypeResolveService
     ?: scopeArgument.resolvedClassArgumentTarget(typeResolver)?.classId
 }
 
+internal fun FirAnnotation.scopeName(session: FirSession): String? {
+  return resolvedScopeClassId()?.shortClassName?.identifier
+    ?: scopeArgument()
+      ?.argument
+      ?.toReference(session)
+      ?.expectAs<FirSimpleNamedReference>()
```
ZacSweers, line 867: nit: let's use `.expectAsOrNull()` since this may not be fully resolved yet in the IDE at the time of writing in the editor

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -113,14 +114,19 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
...
+              val metroContribution = originClass.nestedContributionClassId(scopeClassId)
               getOrPut(scopeClassId, ::mutableSetOf).add(metroContribution)
             }
         }
       }
     }
 
+  private fun FirRegularClassSymbol.nestedContributionClassId(scopeClassId: ClassId): ClassId {
+    return classId.createNestedClassId(
+      Symbols.Names.metroContribution.plus(scopeClassId.shortClassName.identifier)
```
ZacSweers, line 126: I worry this case, however unlikely, would be an issue. What do you think of using `NameAllocator` to allocate names deterministically in this scenario so we avoid that? If we combine that with using a holder annotation to hold the real scope, we should be able to still retrieve the "real" scope key without worrying about what the name of the holding contribution class is.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
```diff
@@ -223,13 +237,19 @@ internal class ContributionsFirGenerator(session: FirSession) :
     if (!owner.hasOrigin(Keys.MetroContributionClassDeclaration)) return emptyList()
     val origin = owner.getContainingClassSymbol() as? FirClassSymbol<*> ?: return emptyList()
     val contributions = findContributions(origin) ?: return emptyList()
+    val scope = owner.name.parseEncodedContributionScope()
```
ZacSweers, line 240: Could we generate the scope annotation onto the contribution somewhere to make this less stringy? Even if it's an internal breadcrumb annotation

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
JoelWilcox: > I worry this case, however unlikely, would be an issue. What do you think of using NameAllocator to allocate names deterministically in this scenario so we avoid that? If we combine that with using a holder annotation to hold the real scope, we should be able to still retrieve the "real" scope key without worrying about what the name of the holding contribution class is.

Not sure if some of the text got cut off but are you thinking of the case where there are multiple scopes in different packages but with the same name? Not familiar with `NameAllocator` so I need to take a look at that but switching to a different naming pattern sounds reasonable to me when coupled with an annotation to store the scope info.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
JoelWilcox: Will give it a shot :+1:, I was thinking we may not be able to copy over the scope parameter to a new annotation if it's not resolved early enough

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
ZacSweers: Yup exactly, same simple name but different packages

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
ZacSweers: I think as long as we use the right predicate setup though it should work. We've already read the scope class ID to generate the name so 🤞

runtime/src/commonMain/kotlin/dev/zacsweers/metro/internal/MetroContribution.kt
ZacSweers, line 1: 👍

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
```diff
@@ -181,7 +227,21 @@ internal class ContributionsFirGenerator(session: FirSession) :
...
+            replaceArgumentMapping(
+              buildAnnotationArgumentMapping {
+                val originalScopeArg =
+                  contributingClassToScopedContributions.getValueIfComputed(owner)?.get(name)
+                    ?: error("Could not find a contribution scope for ${owner.classId}.$name")
+                this.mapping.put("scope".asName(), originalScopeArg)
+              }
+            )
+          }
+        replaceAnnotations(annotations + listOf(metroContributionAnnotation))
```
ZacSweers, line 243: We can also add this in IR if that's easier

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
```diff
@@ -50,6 +61,35 @@ import org.jetbrains.kotlin.name.Name
...
+          .annotationsIn(session, session.classIds.allContributesAnnotations)
+          .toList()
+
+      val contributionNamesToScopeArgs = mutableMapOf<Name, FirGetClassCall?>()
+
+      if (contributionAnnotations.isNotEmpty()) {
+        // We create a contribution class for each scope being contributed to. E.g. if there are
+        // contributions for AppScope and LibScope we'll create $$MetroContribution and
+        // $$MetroContribution2
+        val nameAllocator = NameAllocator(mode = Mode.COUNT)
```
ZacSweers, line 79: 👍

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -67,31 +66,26 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
   private val inCompilationScopesToContributions:
     FirCache<FirSession, Map<ClassId, Set<ClassId>>, TypeResolveService> =
     session.firCachesFactory.createCache { session, typeResolver ->
-      val scopesToContributingClass = mutableMapOf<ClassId, MutableSet<ClassId>>()
```
ZacSweers, line 70: Sorry I kinda kludged some of this on main not realizing it would mess with this. But hopefully it's straightforward, my change was to basically simplify this cache to make it use scope `ClassId`s as the input and the value a `Set<ClassId>`, essentially promoting that `Map<ClassId, Set<ClassId>>` that was here before to be the signature of this FIR cache instead

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
ZacSweers: Ah wait I'm guessing it needs to be done here since here is where we have the tracked context 👍

PR #331, opened 2025-04-18
Improvements for lookup recording to support IC

bnorm:


ZacSweers:
🚀

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -187,6 +188,16 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
         }
         .filterNotTo(mutableSetOf()) { it == StandardClassIds.Nothing }
 
+    for (classId in scopes) {
+      session.lookupTracker?.recordFqNameLookup(
+        Symbols.FqNames.scopeHint(classId),
+        classLikeDeclaration.source,
+        // The class source is the closest we can get to the file source,
+        // and the file path lookup is cached internally.
+        classLikeDeclaration.source,
```
bnorm, line 197: This is the only slightly questionable part. It really should be the file source, but I haven't found a way to get the FirFile for a class. Internally this works though, as it is able to get the path from any kind of source, and the results are cached so it should be fine. If we add more of these lookups, though, we'll want to make sure we use a common "file source" so the cache hit is high.

ZacSweers: I have generally wondered if we should try to cache more things in `IrMetroContext`, maybe another good candidate?

bnorm: This code is all FIR, so caching in IR wouldn't really help.

Also, this is all fine, but I just wanted to note that it's generally not a good idea to use the same source for both parameters. If you keep using a different source element for the `fileSource`, you'll never get a cache hit and you'll explode the cache as well. So pick something as close to the file as you can get, ideally the file obviously, but then try and reuse it with each lookup record so you get cache hits.

ZacSweers: ah oops, I misread where this comment was. Yeah that makes sense, can revisit if we find a good caching spot in the future 👍

PR #343, opened 2025-04-21
Fix rank processing error when the outranked binding is contributed using Metro's ContributesBinding annotation

JoelWilcox:
Previously the rank processing logic made the assumption that both bindings are using Dagger-Anvil's ContributesBinding annotation but that's not necessary when the outranked binding has a default rank (aka no rank).

Hopefully the last rank change 😅


ZacSweers (approved):

JoelWilcox:
- [x] Just noticed there's still an issue for the downstream contribution case, looking into it

JoelWilcox:
> * Just noticed there's still an issue for the downstream contribution case, looking into it

Ended up filing https://youtrack.jetbrains.com/issue/KT-76954/Some-type-arguments-are-not-saved-to-metadata-in-FIR related to the downstream contribution use-case. Basically we're missing the necessary info to fully support this use-case right now. These changes will still fix the original issue for same-compilation use-cases and prevent false-positive errors where rank processing would throw an error in situations it shouldn't (any true errors will still be caught later by duplicate/missing binding checks).

ZacSweers (approved):

PR #344, opened 2025-04-22
Introduce IR dump base test

bnorm:
Couldn't find a good test to convert, but this may be useful if you want to see what is being generated in IR for some example. (Hence the "scratch" test)


compiler-tests/src/test/kotlin/dev/zacsweers/metro/compiler/AbstractIrDumpTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+      defaultDirectives {
+        JVM_TARGET.with(JvmTarget.JVM_11)
+        +FULL_JDK
+        +WITH_STDLIB
+
+        +IGNORE_DEXING // Avoids loading R8 from the classpath.
+
+        -DUMP_IR
+        +DUMP_KT_IR
```
bnorm, line 38: I disabled the default `irElement.dump()` directive (`DUMP_IR`) and enabled the non-default `irElement.dumpKotlinLike()` directive instead (`DUMP_KT_IR`). You can explicitly enable the `DUMP_IR` directive in a test though with `// DUMP_IR` at the top of the file if you want to see the more detailed IR dump.

ZacSweers (approved):

PR #350, opened 2025-04-24
Failing tests to demonstrate issue related to: SingleIn + ContributesBinding + interface naming

japplin:
Failing test for SingleIn not being respected when injected into a wrapper type and that wrapper type is requested via a binding and the interface of that binding starts with a letter which comes after the type marked with SingleIn


ZacSweers:
what a deliciously weird bug

sure enough, AImpl is getting initialized before the scoped B field. Time to debug why
image

ZacSweers:
Continued in #368

PR #354, opened 2025-04-25
Use compiler's addFakeOverrides

kevinguitar:
Found one more issue related to generics, the test fails with the error:
```
Caused by: java.lang.IllegalStateException: Unrecognized type! T (class org.jetbrains.kotlin.ir.symbols.impl.IrTypeParameterSymbolImpl)
	at dev.zacsweers.metro.compiler.ir.IrKt.rawType(ir.kt:146)
	at dev.zacsweers.metro.compiler.ir.BindingKt.injectedClassBindingOrNull(Binding.kt:453)
	at dev.zacsweers.metro.compiler.ir.BindingKt.injectedClassBindingOrNull$default(Binding.kt:446)
	at dev.zacsweers.metro.compiler.ir.BindingGraph.getOrCreateBinding(BindingGraph.kt:189)
	at dev.zacsweers.metro.compiler.ir.BindingGraph.checkCycles(BindingGraph.kt:305)
	at dev.zacsweers.metro.compiler.ir.BindingGraph.validate(BindingGraph.kt:205)
	at dev.zacsweers.metro.compiler.ir.transformers.DependencyGraphTransformer.getOrBuildDependencyGraph(DependencyGraphTransformer.kt:716)
	... 98 more
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 **Unreleased**
 --------------
 
+- **Fix**: Fix generic members inherited from generic supertypes of contributed graphs.
+
 0.3.1
 -----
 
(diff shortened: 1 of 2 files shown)


ZacSweers:
maybe Go was right and generics were a mistake

ZacSweers:
I think this one may be a bug in kotlinc. We generate a declaration in FIR with a supertype of `LoggedInGraph`, but for some reason the generated IR uses the type argument rather than the materialized `Int` type.

```kotlin
@DependencyGraph(scope = LoggedInScope::class, isExtendable = false)
class $$ContributedLoggedInGraph : LoggedInGraph {
  constructor(@Extends exampleGraph: ExampleGraph) /* primary */ {
    super/*Any*/()
    /* <init>() */

  }

  /* fake */ override /* static */ val value: T
    /* fake */ override get(): T

}
```

@bnorm do you think I should file a bug?

ZacSweers (approved):

Issue #365, opened 2025-04-27
Don't suggest moving @Inject to class for Dagger interop


gabrielittner:
### Description

The warning for moving @Inject from an empty constructor to the class shouldn't be given when using Dagger interop or at least not for `javax.inject.Inject` since the annotation itself doesn't support classes as target.
```
w: file:///.../src/main/kotlin/test/Test.kt:30:12 There are no parameters on the @Inject-annotated constructor. Consider moving the annotation to the class instead.
```

### Self-contained Reproducer

```
class Test @javax.inject.Inject constructor()
```

### Metro version

0.3.0-SNAPSHOT

ZacSweers:
This should be an easy one to fix, basically look up the annotation class and its available targets and don't recommend if targets don't include `CLASS`

PR #371, opened 2025-04-29
Don't suggest moving @Inject to class for annotations that don't have CLASS target

kevinguitar:
Resolves https://github.com/ZacSweers/metro/issues/365


ZacSweers:
Thanks!

PR #373, opened 2025-04-29
Fix @ContributesGraphExtension missing binding receiver info for contributed bindings

### LINKED ISSUE #372: ContributesGraphExtension fails to compile when it includes contributions from ContributesBinding
### Description

When a `@ContributesBinding` binding is contributed to a graph extension using `@ContributesGraphExtension`, it results in the following compilation failure:

```
e: java.lang.IllegalStateException: Missing receiver parameter for @Binds function: /* fake */ override /* static */ fun <get-bindAsContributedInterface>(): ContributedInterface

 in class test/ExampleGraph.$$ContributedLoggedInGraph
	at dev.zacsweers.metro.compiler.ir.transformers.DependencyGraphTransformer.createBindingGraph(DependencyGraphTransformer.kt:1108)
	at dev.zacsweers.metro.compiler.ir.transformers.DependencyGraphTransformer.getOrBuildDependencyGraph(DependencyGraphTransformer.kt:706)
	at dev.zacsweers.metro.compiler.ir.transformers.DependencyGraphTransformer.visitClass(DependencyGraphTransformer.kt:213)
```

### Self-contained Reproducer

This occurs with both a single module and multi-module setup. E.g. for the single module case:
```
          abstract class LoggedInScope

          interface ContributedInterface

          @ContributesBinding(LoggedInScope::class)
          @Inject
          class Impl : ContributedInterface
          

          @ContributesGraphExtension(LoggedInScope::class)
          interface LoggedInGraph {
            val contributedInterface: ContributedInterface

            @ContributesGraphExtension.Factory(AppScope::class)
            interface Factory {
              fun createLoggedInGraph(): LoggedInGraph
            }
          }

          @DependencyGraph(scope = AppScope::class, isExtendable = true)
          interface ExampleGraph
```

Can put up a PR to fix shortly

### Metro version

0.3.0-SNAPSHOT

### Context

_No response_

JoelWilcox:
Fixes #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Changelog
 - **Fix**: Fix `@Provides` graph parameters not getting passed on to extended child graphs.
 - **Fix**: Fix qualifiers on bindings not getting seen by extended child graphs.
 - **Fix**: Fix qualifiers getting ignored on accessors from `@Includes` dependencies.
+- **Fix**: Fix `@ContributesGraphExtension` missing binding receiver info for contributed bindings.
 - **Change**: `InstanceFactory` is no longer a value class. This wasn't going to offer much value in practice.
 - **Change**: Change debug reports dir to be per-compilation rather than per-platform.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,6 @@ Changelog
 - **Fix**: Fix `@Provides` graph parameters not getting passed on to extended child graphs.
 - **Fix**: Fix qualifiers on bindings not getting seen by extended child graphs.
 - **Fix**: Fix qualifiers getting ignored on accessors from `@Includes` dependencies.
-- **Fix**: Fix `@ContributesGraphExtension` missing binding receiver info for contributed bindings.
 - **Change**: `InstanceFactory` is no longer a value class. This wasn't going to offer much value in practice.
 - **Change**: Change debug reports dir to be per-compilation rather than per-platform.
 
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributedGraphGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributedGraphGenerator.kt
@@ -34,7 +34,6 @@ import org.jetbrains.kotlin.ir.util.copyParametersFrom
 import org.jetbrains.kotlin.ir.util.createThisReceiverParameter
 import org.jetbrains.kotlin.ir.util.defaultType
 import org.jetbrains.kotlin.ir.util.dumpKotlinLike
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Changelog
 - **Fix**: Fix `@Provides` graph parameters not getting passed on to extended child graphs.
 - **Fix**: Fix qualifiers on bindings not getting seen by extended child graphs.
 - **Fix**: Fix qualifiers getting ignored on accessors from `@Includes` dependencies.
+- **Fix**: Fix `@ContributesGraphExtension` missing binding receiver info for contributed bindings.
```
ZacSweers, line 19: Can omit this one since CGE's aren't released yet

```suggestion
```

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributedGraphGenerator.kt
```diff
@@ -238,6 +237,7 @@ internal class IrContributedGraphGenerator(
               .apply {
                 overriddenSymbols += originalGetter.symbol
                 copyAnnotationsFrom(member.ir)
+                extensionReceiverParameter = originalGetter.extensionReceiverParameter
```
ZacSweers, line 240: Do we need this on the property itself too?

CHANGELOG.md
JoelWilcox: :+1: I was debating if it made sense or not, will remove

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributedGraphGenerator.kt
JoelWilcox: Good question. At the moment I don't have a repro/failure case indicating it is needed. I'm also not sure if there is an equivalent API available for the property itself. I see that we can set `thisReceiver` although I don't think that's the one we want and it blows up a lot of tests when I tried it right now 😁.

ZacSweers (approved):

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributedGraphGenerator.kt
ZacSweers: Can revisit later 🤷. Wish there was just a `IrOverridableDeclaration.overrideIn(targetContainer)` API

JoelWilcox: Yea I agree, would be nice to be able to duplicate methods/properties in their entirety instead of via builders

Issue #364, opened 2025-04-27
Assisted injection codegen fails when factory has super type


gabrielittner:
### Description

We have some factories that all inherit from the same interface that defines the factory method, this fails with the following exception:
```
e: org.jetbrains.kotlin.util.FileAnalysisException: While analysing .../test/src/main/kotlin/test/Test.kt:34:5: org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: Exception in declaration checkers
        at org.jetbrains.kotlin.util.AnalysisExceptionsKt.wrapIntoFileAnalysisExceptionIfNeeded(AnalysisExceptions.kt:57)
        at org.jetbrains.kotlin.fir.FirCliExceptionHandler.handleExceptionOnFileAnalysis(Utils.kt:251)
        at org.jetbrains.kotlin.fir.pipeline.AnalyseKt.runCheckers(analyse.kt:68)
        at org.jetbrains.kotlin.fir.pipeline.FirUtilsKt.resolveAndCheckFir(firUtils.kt:77)
        at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFrontendPipelinePhase.executePhase(JvmFrontendPipelinePhase.kt:167)
        at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFrontendPipelinePhase.executePhase(JvmFrontendPipelinePhase.kt:56)
        at org.jetbrains.kotlin.cli.pipeline.PipelinePhase.phaseBody(PipelinePhase.kt:68)
        at org.jetbrains.kotlin.cli.pipeline.PipelinePhase.phaseBody(PipelinePhase.kt:58)
        at org.jetbrains.kotlin.config.phaser.SimpleNamedCompilerPhase.phaseBody(CompilerPhase.kt:215)
        at org.jetbrains.kotlin.config.phaser.NamedCompilerPhase.invoke(CompilerPhase.kt:111)
        at org.jetbrains.kotlin.backend.common.phaser.CompositePhase.invoke(PhaseBuilders.kt:28)
        at org.jetbrains.kotlin.config.phaser.CompilerPhaseKt.invokeToplevel(CompilerPhase.kt:62)
        at org.jetbrains.kotlin.cli.pipeline.AbstractCliPipeline.runPhasedPipeline(AbstractCliPipeline.kt:106)
[...]

PR #378, opened 2025-05-02
Fix error reporting for missing return type in AssistedFactory

kevinguitar:
Resolves https://github.com/ZacSweers/metro/issues/364

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
@@ -73,7 +73,7 @@ internal object AssistedInjectChecker : FirClassChecker(MppCheckerKind.Common) {
       reporter.reportOn(
         function.source,
         ASSISTED_INJECTION_ERROR,
-        "`@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
+        "Invalid return type: ${targetType.symbol.classId.asSingleFqName()}. `@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
         context,
       )
       return
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
@@ -71,7 +71,7 @@ internal object AssistedInjectChecker : FirClassChecker(MppCheckerKind.Common) {
       }
     if (injectConstructor == null) {
       reporter.reportOn(
-        targetType.source,
+        function.source,
         ASSISTED_INJECTION_ERROR,
         "`@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
         context,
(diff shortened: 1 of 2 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
```diff
@@ -71,7 +71,7 @@ internal object AssistedInjectChecker : FirClassChecker(MppCheckerKind.Common) {
       }
     if (injectConstructor == null) {
       reporter.reportOn(
-        targetType.source,
```
kevinguitar, line 74: The issue is that `targetType.source` could be null if the current FIR session cannot resolve, for example Unit (missing return type).

Changing the report source to the assisted factory function fixed it, I think it's more suitable than `targetType` which might be unpredictable.

Another solution will be splitting the checks, we keep the current `targetType.source` reporting when it's presented, otherwise report the unexpected type when `targetType.source` is missing. Let me know what you prefer.

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/AssistedFactoryTransformerTest.kt
```diff
@@ -235,7 +235,28 @@ class AssistedFactoryTransformerTest : MetroCompilerTest() {
       expectedExitCode = COMPILATION_ERROR,
     ) {
       assertDiagnostics(
-        "e: ExampleClass.kt:6:7 `@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor."
+        "e: ExampleClass.kt:10:7 `@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor."
```
kevinguitar, line 238: Adjusted the report location from `ExampleClass` to `ExampleClassFactory.create`

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
```diff
@@ -71,9 +71,9 @@ internal object AssistedInjectChecker : FirClassChecker(MppCheckerKind.Common) {
...
+        function.source,
         ASSISTED_INJECTION_ERROR,
-        "`@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
+        "Invalid return type ${targetType.symbol.classId.asSingleFqName()}. `@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
```
kevinguitar, line 76: Enhance the error message a bit to mention the return type, similar to Anvil
https://github.com/ZacSweers/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryCodeGen.kt#L129-L130

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
```diff
@@ -73,7 +73,7 @@ internal object AssistedInjectChecker : FirClassChecker(MppCheckerKind.Common) {
       reporter.reportOn(
         function.source,
         ASSISTED_INJECTION_ERROR,
-        "`@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
+        "Invalid return type: ${targetType.symbol.classId.asSingleFqName()}. `@AssistedFactory` target classes must have a single `@Inject`-annotated constructor or be annotated `@Inject` with only a primary constructor.",
```
kevinguitar, line 76: Enhance the error message a bit to mention the return type, similar to Anvil
https://github.com/ZacSweers/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryCodeGen.kt#L129-L130

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/AssistedInjectChecker.kt
ZacSweers: Think this is fine to start 👍

PR #387, opened 2025-05-04
allow AssistedFactory methods to be protected

gabrielittner:
See https://github.com/ZacSweers/metro/issues/364#issuecomment-2849190080

I went with generally supporting this as an option in the `fir.kt` methods. If you thing this goes too far I can also limit it to `singleAbstractFunction` and have a special case only in there.


ZacSweers (approved):
Thanks!

ZacSweers (approved):

PR #393, opened 2025-05-05
Support private constructor injection

### LINKED ISSUE #97: Support private constructor injection
FIR header generation allows us to do this now. Some changes needed:
- Remove visibility check from FIR
- Change logic in IR that depends on the existence of an `@Inject` constructor to be based on the existence of a generated `$$MetroFactory` class. If it's there, we can generate it. This is how the `ProviderFactory` API inside the compiler plugin works for `@Provides` functions today.
- May possibly even want a mode for this to discourage direct access (i.e. encourage separate test constructors? Inclined to recommend that be a separate lint check though.

ZacSweers:
Resolves #97


ZacSweers:
Sort of prohibitively held up by https://youtrack.jetbrains.com/issue/KT-76257. Gonna see if I can work around it by just generating this declaration in FIR for now

ZacSweers:
This can wait until Kotlin 2.2

ZacSweers:
Continued in #472

Issue #379, opened 2025-05-02
Singleton Binding Issue


jimmymorales:
### Description

We've encountered an issue where singleton bindings without qualifiers are not properly maintained as singletons, while singleton bindings with qualifiers work correctly.

## Expected Behavior

Since both platform providers are annotated with `@SingleIn(AppScope::class)`, we expect:
- `platform1 === platform2` (same instance)

## Actual Behavior

- `platform1 !== platform2`  (fails for the unqualified singleton)

## Workaround

Adding a qualifier (such as `@Named`) to the unqualified platform provider resolves the issue:

```kotlin
@Provides
@SingleIn(AppScope::class)
@Named("platform1")  // Adding any qualifier fixes the issue
fun providePlatform(): Platform = getPlatform()
```

And then updating all injection points to use this qualifier:

```kotlin
@Inject
@SingleIn(AppScope::class)
internal class ExampleController(
    @Named("platform1") val platform1: Platform,
    // ...
)
```

## Environment

- Metro DI version: 0.2.0
- Kotlin version: 2.1.20
- Gradle version: 8.9

### Self-contained Reproducer

``` kotlin
interface Platform {
    val name: String
}

internal expect fun getPlatform(): Platform

@MustBeDocumented
@Target(
    AnnotationTarget.FUNCTION,
    AnnotationTarget.FIELD,
    AnnotationTarget.PROPERTY,
    AnnotationTarget.PROPERTY_GETTER,
    AnnotationTarget.CLASS,
    AnnotationTarget.TYPE,
)
@Retention(AnnotationRetention.RUNTIME)
@MapKey
internal annotation class AuthenticatorTypeKey(val value: KClass<out Authenticator.Type>)

@Inject
@SingleIn(AppScope::class)
internal class ExampleController(
    val platform: Platform,
    val authFactory: Authenticator.Factory<Authenticator>
)

@Inject
internal class ExampleDep(val platform: Platform)

interface Authenticator {
    interface Factory<T : Authenticator> {
        fun create(type: Type): T
    }

    sealed class Type {
        data object PhoneNumber : Type()
    }
}

@ContributesBinding(AppScope::class)
@Inject
@SingleIn(AppScope::class)
internal class AuthenticatorFactory(
    private val authenticators: Map<KClass<out Authenticator.Type>, Authenticator>
) : Authenticator.Factory<Authenticator> {
    override fun create(type: Authenticator.Type): Authenticator {
        return authenticators.getValue(type::class)
    }
[...]

PR #403, opened 2025-05-06
Add a failing test for SingleIn when combining Provider and multibindings

asapha:
In the test, `Impl.singleton` has a new instance each time `Impl` is created.

Seems different from the repro provided in #379


ZacSweers:
Just landed a fairly significant rewrite of the graph logic so merged latest main into your branch to see if that fixes it

ZacSweers (approved):
Nice looks like the main changes fixed it 😁

PR #404, opened 2025-05-06
Add more timing logging

ZacSweers:


ZacSweers:
Followups
- Try to incorporate https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:tracing/tracing-driver/
- Rewrite binding graph for the 3rd time

PR #417, opened 2025-05-09
Fix ContributedGraphExtension that extends same interface as parent causes duplicate binding error

gabrielittner:
```
java.lang.AssertionError: Compilation exited with COMPILATION_ERROR but expected OK:
e: file:///var/folders/g7/204fx4ls6pqgcf874nc710540000gp/T/junit12956730234179039896/compilation0/sources/Parent.kt [Metro/DuplicateBinding] Duplicate binding for test.Test
├─ Binding 1: Unknown source location, this may be generated by the compiler.
└─ Here's some additional information we have for the binding:
   ├─ Binding type: Alias
   └─ Binding information: @Provides $$ContributedParentGraph.Test: Test

├─ Binding 2: Unknown source location, this may be generated by the compiler.
└─ Here's some additional information we have for the binding:
   ├─ Binding type: Alias
   └─ Binding information: @Provides ExampleGraph.Test: Test
```


ZacSweers:
This is a sorta tricky one. Basically, right now we add convenience bindings for all graphs' supertypes to the concrete instance. This obviously mucks with that because there are now two bindings. We do this mostly so instance provider lookups have bindings available, but in theory we could require the user to do explicit `@Binds` to get the supertype binding.

A question I have for you would be what you would expect here? I think there's sort of three options. I'm not sure off-hand what dagger does in this scenario. My personal inclination is option 2.

1. Require all explicit bindings. Have the compiler manage finding instances separately

```kotlin
@ContributesGraphExtension(Parent::class, isExtendable = true)
interface ParentGraph : Test {
  @Binds val ParentGraph.bind: Test // This one wins. Or define it in ExampleGraph and then that one ones
  @ContributesGraphExtension.Factory(AppScope::class)
  interface Factory {
    fun create(): ParentGraph
  }
}

@DependencyGraph(scope = AppScope::class, isExtendable = true)
interface ExampleGraph : Test
```

2. Keep convenience bindings, make the child's binding override the parent's conflicting one. No change to your source test in that case.

```kotlin
@ContributesGraphExtension(Parent::class, isExtendable = true)
interface ParentGraph : Test {
  @ContributesGraphExtension.Factory(AppScope::class)
  interface Factory {
    fun create(): ParentGraph
  }
}

@DependencyGraph(scope = AppScope::class, isExtendable = true)
interface ExampleGraph : Test
```

We could add an extra layer to this like `@Binds @ReplaceMatchingParentBinding val ParentGraph.bind: Test` to make the user explicitly indicate they are overriding.

**3. Hybrid solution of 1 and 2**

Require explicit bindings but allow children to override parents' duplicate one.

ZacSweers:
My internet sleuthing suggests dagger also errors in this case, which might require option 1 in this case then

gabrielittner:
Dagger does also fail on duplicate bindings (at least for subcomponents), the difference is that Dagger doesn't automatically add a binding for each interface that the component extends. That and consistency with other types of bindings (like `@Binds`, contributed bindings, provided objects) would make me lean towards option 1. However I'm not sure how big of a use cases having the interfaces bound automatically (outside of the internal metro usages you mention) is for others.

ZacSweers:
It's really more for metro (because we subclass the graph type that the provider might belong to), so I think I'm ok with doing that. Will require a bit of work to tweak how it tracks these available bindings to make them not participate in the internal binding graph by default

ZacSweers:
Worked together an implementation I'm happy with

PR #418, opened 2025-05-09
Add KSP to compiler test framework

bnorm:
There's still some work here, but I got two tests passing!


compiler-tests/src/test/data/box/interop/dagger/DaggerFactoryClassCanBeLoaded.kt
```diff
@@ -0,0 +1,24 @@
+// ENABLE_DAGGER_KSP
+
+// FILE: ExampleClass.java
+import javax.inject.Inject;
+
+public class ExampleClass {
+  @Inject public ExampleClass() {
+
+  }
+}
```
bnorm, line 10: One weird issue I found is that:

If `package test;` is specified, Kotlin compiles `ExampleGraph` without needing an import, but then fails at runtime because it thinks that `ExampleClass` is in the root package. `ExampleGraph_Factory` is correctly in the `test` package and Kotlin compiles against it correctly.

If the file name is changed to `test/ExampleClass.java`, then an import is needed so things seem to be on the right track, but then a `could not find generated factory for 'test.ExampleClass' in upstream module where it's defined. Run the Metro compiler over that module too.` error is reported and the test fails at runtime.
```
java.lang.IllegalStateException: Never called
	at ExampleGraph$$$MetroGraph.getExampleClass(ExampleGraph.kt:19)
```

I'm not sure if this is something wrong with the compiler test framework, configuration of Metro, something I'm doing wrong in the additional source provider, a bug in Metro, a weird alignment of the planets, or because it's a Friday. For now, it works without the package, so I left it without the package.

ZacSweers:
Iterated on your branch a little bit, but stumped as to why the daggerInterop runtime symbols are missing in the failing tests 🤔

bnorm:
> why the daggerInterop runtime symbols are missing in the failing tests 🤔

The only test failing for me is the `scratch.kt` test, which is using the wrong directive name. The file should ideally never by updated since it's more of a developer utility than an actual test. So I shouldn't have committed a change to that file anyways and it can be reverted.

bnorm:
> why the daggerInterop runtime symbols are missing in the failing tests 🤔

Also, it's because I forgot to add a `dependsOn(daggerInteropClasspath)` to the compiler-test task.... 🫣

ZacSweers (approved):

PR #422, opened 2025-05-11
Avoid unnecessary StringBuilder allocations

jzbrooks:
Noticed a few situations where a string builder was allocated for interpolation in the context of an existing string builder while reading through the project today.

A number of these are in error paths, so hopefully not hot, but still seems worthwhile to me.


ZacSweers:
Thanks! Sorry I had some large in-flight changes that just landed in main so there are some conflicts, do you mind updating?

jzbrooks:
No worries. Those changes look more important than this 😄 

Done!

PR #423, opened 2025-05-11
Binding graph v3: tarjan's algorithm + topological sorting

### LINKED ISSUE #80: Improve dependency graph field generation
Dependency graph validation could be improved! Currently lattice does validation in one pass and then separately collects bindings that need provider fields. These provider fields are then sorted by codependency to determine init order. This is performant enough but is a little messy. I'd love to streamline this with a little inspiration from dagger, where we track "field generators" while validating the graph and "visit" those generators during our recursion.

Generators would have an internal state of `uninitialized`, `initializing`, `deferred`, and `initialized`. Every time we visit a type, we get or create a generator and "visit" it. If a generator is reentrantly visited while initializing, it means we have a cycle. If it's a valid cycle, we mark it as `deferred` and these types will need to use `DelegateFactory`.

At the end, we collect generators. There may be some duplications from each dfs pass, and in those cases we take either the one that uses delegate factory or any. Generators that result in fields in the end must be one of
- Scoped (and wrapped in DoubleCheck)
- Backed by a `DelegateFactory`
- Used by multiple dependencies

Finally, any deferred fields will also need a corresponding initializer in the constructor.

If we keep insertion order of these generators as they're visited, my hope is we can infer the init order directly from that (i.e. reverse the order they were added).

ZacSweers:
This PR reworks binding graph validation + initialization ordering + deferred types in one fell swoop with tarjan's algorithm + topological sorting (h/t @swankjesse).

Couple other things along the way
- Effectively resolves #80
- Reworked multibindings to use a clever trick from Guice to make each into* provider have a synthetic `@MultibindingElement` qualifier with a deterministic binding ID + deterministic-but-unique element ID to allow them to participate normally in the dependency graph's resolution.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/DependencyGraphTransformer.kt
```diff
@@ -1577,23 +1524,7 @@ internal class DependencyGraphTransformer(
...
-                  b.dependsOn(a) -> -1
-                  // Otherwise order doesn't matter, fall back to just type order for idempotence
-                  else -> a.compareTo(b)
-                }
-              }
-            }
-            .map { bindingDependencies.getValue(it) }
-            .distinct()
-        }
+        sealResult.sortedKeys.mapNotNull { bindingDependencies[it] }.distinctBy { it.typeKey }
```
ZacSweers, line 1527: chefskiss.png

Issue #377, opened 2025-05-01
Bindings scoped to parent graphs but unused in parent graphs are scoped to extensions instead 


asapha:
### Description

If we don't help it, Metro doesn't infer that it's fine to use an `AppScope` dependency in a `LoggedInScope` one.

```kotlin
[Metro/IncompatiblyScopedBindings] test.ExampleGraph.$$ContributedLoggedInGraph (scopes '@SingleIn(LoggedInScope::class)') may not reference bindings from different scopes:
    test.Dependency (scoped to '@SingleIn(AppScope::class)')
    test.Dependency is injected at
        [test.ExampleGraph.$$ContributedLoggedInGraph] test.ChildDependency(…, dep)
    test.ChildDependency is requested at
        [test.ExampleGraph.$$ContributedLoggedInGraph] test.LoggedInGraph#childDependency
```

### Self-contained Reproducer

```kotlin
sealed interface LoggedInScope

@Inject @SingleIn(AppScope::class) class Dependency
@Inject @SingleIn(LoggedInScope::class) class ChildDependency(val dep: Dependency)

@DependencyGraph(scope = AppScope::class, isExtendable = true)
interface ExampleGraph {
  // Works if added explicitly like this
  // val dependency: Dependency
}

@ContributesGraphExtension(LoggedInScope::class)
interface LoggedInGraph {
    val childDependency: ChildDependency

    @ContributesGraphExtension.Factory(AppScope::class)
    interface Factory {
        fun createLoggedInGraph(): LoggedInGraph
    }
}
```

### Metro version

0.3.0-SNAPSHOT

### Context

Tried to migrate a project using `kotlin-inject-anvil`

ZacSweers:
hmm, this is indeed a tricky one. Does this scenario work in dagger? Currently Metro graphs don't create fields for scoped constructor-injected classes that aren't actually referenced by anything in the graph. If this works in dagger, I imagine it's because they process the entire parent + subcomponent together in one pass, but that's not an option we immediately have available here.

ZacSweers:
I'm inclined to continue making this an error but with a more descriptive message. Basically - upon encountering a scope mismatch, check if the scope matches any extended parents and just say the parent should declare it somewhere. Unfortunately I'm not sure of a better solution than that for now

kevinguitar:
> Does this scenario work in dagger?

[...]

PR #432, opened 2025-05-13
Enable graph extensions to access scoped constructor-injected classes from parent graphs even if the parent doesn't have an explicit accessor

JoelWilcox:
Fixes the issue originally reported in https://github.com/ZacSweers/metro/issues/377

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/DependencyGraphTransformer.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/DependencyGraphTransformer.kt
@@ -2200,56 +2200,8 @@ internal class DependencyGraphTransformer(
 
     val bindingScope = binding.scope
 
-    // Check scoping compatibility
     // TODO FIR error?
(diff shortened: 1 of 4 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/DependencyGraphTransformer.kt
```diff
@@ -2352,6 +2304,61 @@ internal class DependencyGraphTransformer(
...
   }
 
+  private fun checkScopingCompatibility(
+    node: DependencyGraphNode,
+    binding: Binding,
+    bindingStack: IrBindingStack,
+  ) {
+    val bindingScope = binding.scope ?: return
+
+    if (node.scopes.isEmpty() || !node.hasAccessToScope(bindingScope)) {
```
JoelWilcox, line 2314: This is the notable logic change, the rest of the method should be effectively the same, just pulled into a helper.

This used to check if the bindingScope was in the node's scopes, and now it checks the node's scopes and its parents' scopes

PR #447, opened 2025-05-19
Move more cycles tests to compiler-tests

ZacSweers:


compiler-tests/src/test/data/diagnostic/cycles/CyclesShouldFailAcrossMultipleGraphs.kt
```diff
@@ -0,0 +1,29 @@
+// RENDER_DIAGNOSTICS_FULL_TEXT
```
bnorm: We should probably start using `RENDER_ALL_DIAGNOSTICS_FULL_TEXT` instead, as that should render IR diagnostics as well in the same file. Maybe even configure it as a default directive?

ZacSweers: Seems reasonable, though doesn't capture the errors reported in IR :/

PR #461, opened 2025-05-24
Enable child graphs to depend on parent-scoped dependencies that are unused and not referenced in the parent scope

JoelWilcox:
- Applies to both contributed bindings and scoped constructor-injected classes without contributions.
- Introduces a new Metro option `enableInjectConstructorHints` which determines if we should generate hints for classes with scoped `@Inject` constructors. By default this option is disabled.
- Fixes https://github.com/ZacSweers/metro/issues/377.


JoelWilcox:
Closing in favor of a pure IR implementation after discussing it with Zac: https://github.com/ZacSweers/metro/pull/576

PR #490, opened 2025-05-27
[Documentation] Reword aggregation @ContributesBinding section

FilippoVigani:
### Description

This PR updates the documentation with a small rewording on the `@ContributesBinding` section by stating that you MUST specify a binding if the class has multiple supertypes.

### Rationale
By reading that section you could assume that specifying the bound supertype was an option but not required in case of multiple supertypes. 
However, if you try to do that you correctly will get a compilation error:
```kotlin
`@ContributesBinding`-annotated class @dev.zacsweers.metro.ContributesBinding doesn't declare an explicit `bindingType` but has multiple supertypes. You must define an explicit bound type in this scenario.
```

By rewording the section, we state that in case of multiple supertypes, you must specify the bound supertype.

### DIFF
--- a/docs/aggregation.md
+++ b/docs/aggregation.md
@@ -91,7 +91,7 @@ For simple cases where there is a single supertype, that type is implicitly used
 class CacheImpl(...) : Cache
 ```
 
-For classes with multiple supertypes or advanced cases where you want to bind an ancestor type, you can explicitly define this via `binding` parameter.
+For advanced cases where you want to bind an ancestor type, you can explicitly define this via `binding` parameter. For example, if your class has multiple supertypes, you must define an explicit bound type for it.
 
 ```kotlin
 @Named("cache")


ZacSweers:
I don't think the current wording is ambiguous, but I do think that the proposed is less clear as it implies that multiple supertypes are an advanced case (they're not, just one case where it needs to be specified)?

FilippoVigani:
My only issue is with `can` in 
> For classes with multiple supertypes [...] you can explicitly define this via `binding` parameter.

To me, this doesn't imply that not specifying the binding would cause a compilation error. So I tested it locally just to make sure that was the behavior, instead of for example binding each individual supertype, or binding the first one, etc.

Just my two cents. Feel free to reword it differently or leave it as is, not a big deal.

ZacSweers:
Think I wanna keep the current wording as this hasn't come up anywhere else as an issue, but thank you anyway!

PR #495, opened 2025-05-28
Support generic providers in Graph Factory

kevinguitar:
It's me with generic issues again 😅 

The test reproduces the following stacktrace:
```
e: org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: Not supported: class org.jetbrains.kotlin.fir.types.impl.ConeTypeParameterTypeImpl
	at org.jetbrains.kotlin.fir.types.TypeUtilsKt.withArguments$error(TypeUtils.kt:1121)
	at org.jetbrains.kotlin.fir.types.TypeUtilsKt.withArguments(TypeUtils.kt:172)
	at dev.zacsweers.metro.compiler.fir.generators.DependencyGraphFirGenerator.generateConstructors$lambda$8$lambda$7$lambda$6(DependencyGraphFirGenerator.kt:347)
	at org.jetbrains.kotlin.fir.plugin.FunctionBuildingContext.generateValueParameter(FunctionBuildingContext.kt:86)
	at org.jetbrains.kotlin.fir.plugin.ConstructorBuildingContext.build$lambda$5(ConstructorBuildingContext.kt:62)
	at org.jetbrains.kotlin.fir.plugin.ConstructorBuildingContext.build(ConstructorBuildingContext.kt:185)
	at org.jetbrains.kotlin.fir.plugin.ConstructorBuildingContextKt.createConstructor(ConstructorBuildingContext.kt:133)
	at dev.zacsweers.metro.compiler.fir.generators.DependencyGraphFirGenerator.generateConstructors(DependencyGraphFirGenerator.kt:333)
```

My question is, should this scenario be supported? If so, I can look into a fix 🙏 

Resolves https://github.com/ZacSweers/metro/issues/504

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
@@ -353,7 +353,7 @@ internal class DependencyGraphFirGenerator(session: FirSession) :
                     if (valueParameterSymbol.resolvedReturnType is ConeTypeParameterType) {
                       valueParameterSymbol.resolveReturnTypeFrom(
                         typeOwner = creator.classSymbol,
-                        session = session
+                        session = session,
                       )
                     } else {
                       valueParameterSymbol.resolvedReturnType.withArguments(
@@ -424,10 +424,14 @@ internal class DependencyGraphFirGenerator(session: FirSession) :
                   if (parameter.resolvedReturnType is ConeTypeParameterType) {
                     val creator =
                       graphObject(context.owner.requireContainingClassSymbol())
-                        ?.findCreator(session, "generateConstructors for ${context.owner.classId}", ::log)
+                        ?.findCreator(
+                          session,
+                          "generateConstructors for ${context.owner.classId}",
+                          ::log,
+                        )
                     parameter.resolveReturnTypeFrom(
                       typeOwner = creator?.classSymbol!!,
-                      session = session
+                      session = session,
                     )
                   } else {
                     parameter.resolvedReturnType.withArguments(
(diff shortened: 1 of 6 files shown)


ZacSweers:
If it works in dagger and KI then I'd like to support it here too 👍

kevinguitar:
It works in Dagger, not sure about KI. I'll try and figure out a fix then 👍

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/TestUtils.kt
```diff
@@ -422,7 +422,7 @@ fun Class<*>.createGraphWithNoArgs(): Any {
 fun Class<*>.createGraphViaFactory(vararg args: Any): Any {
   val factoryInstance = invokeGraphFactory()
   return factoryInstance.javaClass.declaredMethods
-    .single { it.name == Symbols.StringNames.CREATE }
+    .first { it.name == Symbols.StringNames.CREATE }
```
kevinguitar, line 425: Found out that if the graph creation function lives in a super type, there will be multiple "create" functions in the declared methods. I changed `single` to `first` to unblock the test for now

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -403,9 +421,23 @@ internal class DependencyGraphFirGenerator(session: FirSession) :
...
+                        )
+                    parameter.resolveReturnTypeFrom(
+                      typeOwner = creator?.classSymbol!!,
+                      session = session,
+                    )
+                  } else {
+                    parameter.resolvedReturnType.withArguments(
+                      it.mapToArray(FirTypeParameterRef::toConeType)
+                    )
+                  }
```
kevinguitar, line 440: This part can be improved probably, but not sure what's the ideal way to do it. The problem is we need `creator` to resolve the actual generic type, I tried to make it as lazy as possible here

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -339,9 +350,16 @@ internal class DependencyGraphFirGenerator(session: FirSession) :
                   name = valueParameterSymbol.name,
                   key = Keys.RegularParameter,
                   typeProvider = {
```
bnorm, line 352: The use of `typeProvider` here seems pointless since you can't any type parameters to the constructor. `typeProvider` isn't actually that lazy and will run before the `createConstructor` function exits. It's used when you need to add value parameters / return type to depend on type parameters of a _**function**_. So the whole `withArguments` mapping is completely useless.

Same goes for generating functions for both the value parameters and return type. Both values should just be provided directly and don't need the `typeProvider` style.

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/TestUtils.kt
kevinguitar: yep, I'll do that

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
kevinguitar: > So the whole withArguments mapping is completely useless.

Gotcha! Makes sense

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -548,3 +580,46 @@ internal class DependencyGraphFirGenerator(session: FirSession) :
     }
   }
 }
+
+private fun FirValueParameterSymbol.resolveReturnTypeFrom(
```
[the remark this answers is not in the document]
ZacSweers: Let's try to remove all of them in this file. If we cannot resolve a return type then just exit processing that declaration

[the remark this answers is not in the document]
kevinguitar: Done, this function doesn't throw anymore

ZacSweers (approved):
Made a couple tweaks, thanks!

PR #496, opened 2025-05-28
Failing test for an IC issue on contributed class visibility changes

kevinguitar:
Spotted a small nasty IC issue during migration, created a failing test for it.


ZacSweers:
Thanks for the repro. I wonder if this may also be fixed by #472 along the way since it adds a direct IC lookup to the class that wasn't previously there. I'm going to mark this test as ignored for now so I can merge it and then try to enable it in that branch

ZacSweers:
Finding this one tricky to test because the only way to connect a debugger in my experience is with `-Pkotlin.compiler.execution.strategy=in-process`, but that disables IC 🥲

ZacSweers:
weird CI failure, I wonder if the cache in github actions is busted

ZacSweers:
yeah it's [this](https://github.com/gradle/actions/issues/72). Merging past the failure

ZacSweers:
Guh, that other PR doesn't fix it unfortunately. Will bring this up with compiler folks

Issue #104, opened 2025-01-07
Improve generics handling in FIR


ZacSweers:
Much of the generics handling in FIR is best-effort and tbh a bit hacky. Digging around in FIR sources, the class to use to handle substitutions is `ConeSubstitutor`. Unfortunately it's undocumented. Need to ask around a bit. Particularly when it comes to generics from inherited supertypes, like generic fields.

ZacSweers:
Pointer from kotlin-lang: https://kotlinlang.slack.com/archives/C7L3JB43G/p1736656372131399?thread_ts=1736529087.849159&cid=C7L3JB43G

> While it's not documentation, Kotlin Serialization does this for generic overrides: [https://github.com/JetBrains/kotlin/blob/68f050338875bd9ee1f8e754163949fb2f49f732/[…]/serialization/compiler/fir/SerializationFirResolveExtension.kt](https://github.com/JetBrains/kotlin/blob/68f050338875bd9ee1f8e754163949fb2f49f732/plugins/kotlinx-serialization/kotlinx-serialization.k2/src/org/jetbrains/kotlinx/serialization/compiler/fir/SerializationFirResolveExtension.kt#L204-L213). This may serve as a good example for your use case, but probably not a good example for `ConeSubstitutor`.

ZacSweers:
Another place to check is validating graph creator return types

ZacSweers:
I'm more or less ok with how this is implemented now. There might still be some untested areas around graph factories but can cross that bridge when we get there

PR #510, opened 2025-05-29
Use unsubstituted scope more

ZacSweers:
Helps with #104. Not marking it as resolved just yet but close

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/firFactoryUtil.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/firFactoryUtil.kt
@@ -233,6 +233,12 @@ internal fun FirExtension.buildNewInstanceFunction(
 }
 
 internal fun FirClassSymbol<*>.findSamFunction(session: FirSession): FirFunctionSymbol<*>? {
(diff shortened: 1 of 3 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/firFactoryUtil.kt
```diff
@@ -241,5 +247,5 @@ internal fun FirClassSymbol<*>.findSamFunction(session: FirSession): FirFunction
       memberRequiredPhase = null,
     )
   if (scope.collectAllProperties().any { it.isAbstract }) return null
```
bnorm, line 249: This check should only take place when finding the SAM function, not when collecting all abstract functions. Might be why your test data changed.

ZacSweers: Doesn't seem like it, the test looked broken before to me so I think the updated version is correct. Good point to not exit early here though

ZacSweers: https://github.com/ZacSweers/metro/pull/517

PR #518, opened 2025-05-31
Deterministic IrType.render + sorted sets/maps again

ZacSweers:
This PR attempts to unrevert #515 and also improve (Contextual)IrTypeKey rendering by using a single, deterministic render. The hope is that the latter resolves the issues with the former

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog
 **Unreleased**
 --------------
 
+- **Enhancement:** Improve graph validation performance by avoiding unnecessary intermediate sorts (again).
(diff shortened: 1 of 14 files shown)


ZacSweers:
Got confirmation that sorts the issue. Wanna see if I can make a regression test knowing this was the underlying issue now

ZacSweers:
I'm still unable to make a repro test for this, it's something to do with how the previous impl would include `@JvmSuppressWildcards` annotations in its render and I suspect some wires were crossed due to IrContextualTypeKey stripping those. Proceeding with this though since it locks it all down now

PR #526, opened 2025-06-02
Failing test for ContributesGraphExtension missing binding when using @Includes in parent

kevinguitar:


ZacSweers (approved):
Thanks!

Sidenote would you mind sending future repro tests in the `:compiler-tests` module?

kevinguitar:
Ahh yes! Noted 👍

PR #541, opened 2025-06-05
Failing test for injecting constructor parameters with generics

kevinguitar:
Ran into another generic issue with constructor injection while migrating from Dagger


ZacSweers:
fyi it'd be helpful if you pasted the error too so I don't need to build it locally to start thinking about it :). CI doesn't print detailed errors for failures currently

kevinguitar:
Noted, will do!

ZacSweers:
Continued in #549

PR #549, opened 2025-06-06
Fix generic constructor-injected parameters

ZacSweers:
Continuing from #541


ZacSweers:
Member injection is still in rough shape, and might need some larger reworking of how we model this in IR. Some scratch notes

```kotlin
// TODO
//  - Member injectors (especially deep ancestors)
//  - Member injection (especially deep ancestors)
//    - Setters, functions, properties
@Inject
class ExampleClass<T : Any> {
  @Inject lateinit var value: T
  @Inject lateinit var values: List<T>
  @Inject lateinit var mapValues: Map<T, List<T>>
}

@DependencyGraph
interface AppGraph {
  val exampleClass: ExampleClass<Int>

  @Provides fun provideInt(): Int = 3
  @Provides fun provideInts(): List<Int> = listOf(3)
  @Provides fun provideIntMap(int: Int, ints: List<Int>): Map<Int, List<Int>> = mapOf(3 to ints)
}

fun box(): String {
  val exampleClass = graph.exampleClass
  assertEquals(exampleClass.value, 3)
  assertEquals(exampleClass.values, listOf(3))
  assertEquals(exampleClass.mapValues, mapOf(3 to listOf(3)))
  return "OK"
}
```

PR #552, opened 2025-06-07
Failing test: Includes a deep inherited interface results in duplicated bindings

kevinguitar:
Found an issue that was introduced during https://github.com/ZacSweers/metro/pull/547

The test failed with duplicated bindings:

```
file:///Users/kevin/Documents/metro/compiler-tests/src/test/data/box/dependencygraph/IncludesDeepInheritedInterfacesWork.kt:3:1: error: [Metro/DuplicateBinding] Duplicate binding for kotlin.Int
├─ Binding 1: file:///IncludesDeepInheritedInterfacesWork.kt:13:1 
├─ Binding 2: file:///IncludesDeepInheritedInterfacesWork.kt:19:3 
```


ZacSweers:
Introduced as in it worked prior to that commit?

ZacSweers (approved):
Issue was that it misses the root one if it's not defined in the intermediate one

PR #559, opened 2025-06-09
Track reachable nodes and omit unused bindings

### LINKED ISSUE #33: Report errors for unused graph dependencies or graph bindsinstance deps
Dagger reports something like this on the factory

```
/ExampleComponent.java:25: error: @Component.Factory method has parameters for modules or components that aren't required: [org.jetbrains.annotations.NotNull com.slack.circuit.star.CharSequenceComponent component]
    public static abstract interface Factory {
```

ZacSweers:
Effectively resolves #33 along the way by offering a diagnostic. No good option for warning/failing on unused at the moment for a few reasons (extendable graphs may expose them, extension graphs may inherit a bunch they don't use, etc)

### DIFF
--- a/benchmark/generate_performance_summary.sh
+++ b/benchmark/generate_performance_summary.sh
@@ -72,13 +72,11 @@ calculate_percentage() {
     echo "scale=0; (($value - $baseline) * 100) / $baseline" | bc
 }
 
-# Function to generate performance table for a test type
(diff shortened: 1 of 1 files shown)


ZacSweers:
Gonna need to retarget this onto the kotlin 2.2 branch as this requires some fixes to member injection validation there

PR #576, opened 2025-06-11
Enable child graphs to depend on parent-scoped dependencies that are unused in the parent scope (IR)

JoelWilcox:
- More specifically, this enables the behavior for scoped `@Inject` classes that have no contributions (e.g. with `@ContributesBinding`).
- This behavior is enabled by a new Metro option `enableScopedInjectClassHints` which determines if we should generate hints for classes annotated with `@Inject` and at least one scope annotation. By default this option is disabled.
- Extracts `ClassBindingLookup` from `IrBindingGraph` and most of the hint generation from `ContributionHintIrTransformer` to both be reusable elsewhere.
- Fixes ZacSweers#377.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ClassBindingLookup.kt
```diff
@@ -0,0 +1,115 @@
...
+import dev.zacsweers.metro.compiler.ir.parameters.parameters
+import dev.zacsweers.metro.compiler.ir.transformers.MembersInjectorTransformer.MemberInjectClass
+import dev.zacsweers.metro.compiler.metroAnnotations
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.types.IrSimpleType
+import org.jetbrains.kotlin.ir.util.classIdOrFail
+import org.jetbrains.kotlin.ir.util.getSimpleFunction
+import org.jetbrains.kotlin.ir.util.isObject
+
+internal class ClassBindingLookup(
```
ZacSweers, line 16: Just confirming - no changes in this one right? Just moved it to top-level?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/MetroIrGenerationExtension.kt
```diff
@@ -45,10 +46,23 @@ public class MetroIrGenerationExtension(
             moduleFragment.transform(ContributionBindsFunctionsIrTransformer(context), null)
           }
 
+          val hintGenerator = LookupHintGenerator(context, moduleFragment)
+          // First.6 - transform scoped @Inject classes to add their hints
```
ZacSweers: > First.6

😅

The first.5 was supposed to be like, "first and a half"

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/MetroIrGenerationExtension.kt
```diff
@@ -45,10 +46,23 @@ public class MetroIrGenerationExtension(
             moduleFragment.transform(ContributionBindsFunctionsIrTransformer(context), null)
           }
 
+          val hintGenerator = LookupHintGenerator(context, moduleFragment)
```
ZacSweers: The order of events here is a little different than I was thinking. You can continue generating hints in the existing hint generator (they're not for in-round use), the only thing you need to do here is _lookup_ hints from the classpath, which doesn't need to be a transformer. Instead it can just be a visitor like the `IrContributionData` is above. I would actually expect this lookup to be merged into that collection, was there a reason you wanted it to be separate?

(TL;DR this adds another pass in the compiler but I think we can do it within the existing passes)

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/LookupHintGenerator.kt
```diff
@@ -0,0 +1,135 @@
...
+ * Note that the generated name may take other forms determined by the caller of [generateHint].
+ *
+ * Importantly, we also add these generated functions to metadata via
+ * [IrGeneratedDeclarationsRegistrar.registerFunctionAsMetadataVisible], which ensures they are
+ * visible to downstream compilations.
+ *
+ * File creation is on a little big of shaky ground, but necessary for this to work. More
+ * explanation can be found below.
+ */
+internal class LookupHintGenerator(context: IrMetroContext, val moduleFragment: IrModuleFragment) :
```
ZacSweers: nit: let's just call it `HintGenerator`?

gradle-plugin/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
   functionalTestImplementation(libs.kotlin.stdlib)
   functionalTestImplementation(libs.testkit.support)
   functionalTestImplementation(libs.testkit.truth)
+  functionalTestImplementation(libs.kotlin.compilerEmbeddable)
```
ZacSweers: What was this needed for?

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/ContributesGraphExtensionTest.kt
```diff
@@ -1186,6 +1186,284 @@ class ContributesGraphExtensionTest : MetroCompilerTest() {
     )
   }
 
+  @Test
+  fun `a scoped inject class can be accessed in a child graph without an explicit accessor when associated hints are enabled`() {
```
ZacSweers, line 1190: Doesn't have to be this PR but in the future let's write these as box tests

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ClassBindingLookup.kt
JoelWilcox: That's right -- this is just moved

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/MetroIrGenerationExtension.kt
JoelWilcox: Oops lol meant to come back and update this, I can remove that bit

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/MetroIrGenerationExtension.kt
JoelWilcox: Good catch, I didn't realize that this is adding another pass compared to where ContributionHintIrTransformer gets invoked. My original thinking was that the use-case is distinct enough that it made sense to keep the scoped inject hint handling separate while still reusing any common code.

gradle-plugin/build.gradle.kts
JoelWilcox: Ahh I originally added this with an earlier iteration of the new IC test when I was instantiating the contributed graph inside of the test body (rather than through the generated main() source function) and it was needed for something as part of that. Just verified it's no longer necessary and removed

PR #584, opened 2025-06-12
Implement `FirAccessorOverrideStatusTransformer`

kevinguitar:
This is similar to https://github.com/ZacSweers/metro/pull/547. We have a use case internally, where a contributed `ServiceProvider` has the same accessors as the graph bindings, looks like we need to dedupe here as well.

The error I'm getting:
```
VIRTUAL_MEMBER_HIDDEN: 'int' hides member of supertype 'ServiceProvider' and needs an 'override' modifier. at OverrideCompatibleGraphBindingAccessors.kt:(285,288)
java.lang.IllegalStateException: VIRTUAL_MEMBER_HIDDEN: 'int' hides member of supertype 'ServiceProvider' and needs an 'override' modifier. at OverrideCompatibleGraphBindingAccessors.kt:(285,288)
	at org.jetbrains.kotlin.test.backend.handlers.NoFirCompilationErrorsHandler.processModule(NoFirCompilationErrorsHandler.kt:55)
	at org.jetbrains.kotlin.test.backend.handlers.NoFirCompilationErrorsHandler.processModule(NoFirCompilationErrorsHandler.kt:26)
	at org.jetbrains.kotlin.test.TestStep$HandlersStep.processModule(TestStep.kt:78)
	at org.jetbrains.kotlin.test.TestStep$HandlersStep.processModule(TestStep.kt:56)
	at org.jetbrains.kotlin.test.TestRunner.processModule(TestRunner.kt:205)
	at org.jetbrains.kotlin.test.TestRunner.hackyProcessModule(TestRunner.kt:196)
	at org.jetbrains.kotlin.test.TestRunner.processModule(TestRunner.kt:140)
	at org.jetbrains.kotlin.test.TestRunner.runTestPipeline(TestRunner.kt:91)
```

Note: It's also failing on 2.2 branch


kevinguitar:
If you're aligned, I'd love to try to contribute a fix this weekend, it will be very helpful to migrate our use cases.

The solution looks to me that while generating graph members, we need to check if any `@Include` supertype has override-compatible properties/ functions, if so, we add an override modifier. If it sounds right, it would be nice if you could point me to a file that does this 🙏

ZacSweers:
Makes sense to me 👍

ZacSweers:
FYI this might be slightly tricky as it looks like this fails in FIR, so you'd need to implement a status transformer (see `FirProvidesStatusTransformer`). It might be worth a feature request to youtrack for FIR2IR to handle this automatically when contributing additional supertypes.

Implementing this could be a bit tricky - basically we would need to traverse all members coming in through contributed supertypes, which is gonna be pretty expensive. I'm also not sure off-hand how to directly transform a declaration in FIR to add overridden symbols 🤔 

In the meantime, as a workaround, you _can_ just add `override` to your source declaration and it will work, even if the IDE maybe doesn't realize it

ZacSweers:
I tinkered with it a bit if you want to take this on

```kotlin
// Copyright (C) 2025 Zac Sweers
// SPDX-License-Identifier: Apache-2.0
package dev.zacsweers.metro.compiler.fir

import dev.zacsweers.metro.compiler.Symbols
import dev.zacsweers.metro.compiler.metroAnnotations
import org.jetbrains.kotlin.descriptors.isInterface
import org.jetbrains.kotlin.fir.FirSession
import org.jetbrains.kotlin.fir.analysis.checkers.classKind
import org.jetbrains.kotlin.fir.analysis.checkers.getContainingClassSymbol
import org.jetbrains.kotlin.fir.copy
import org.jetbrains.kotlin.fir.declarations.FirCallableDeclaration
import org.jetbrains.kotlin.fir.declarations.FirDeclaration
import org.jetbrains.kotlin.fir.declarations.FirDeclarationStatus
import org.jetbrains.kotlin.fir.declarations.FirProperty
import org.jetbrains.kotlin.fir.declarations.FirSimpleFunction
import org.jetbrains.kotlin.fir.declarations.utils.hasBody
import org.jetbrains.kotlin.fir.declarations.utils.isAbstract
import org.jetbrains.kotlin.fir.declarations.utils.isOverride
import org.jetbrains.kotlin.fir.extensions.FirDeclarationPredicateRegistrar
import org.jetbrains.kotlin.fir.extensions.FirStatusTransformerExtension
import org.jetbrains.kotlin.fir.extensions.predicateBasedProvider
import org.jetbrains.kotlin.fir.resolve.getSuperTypes
import org.jetbrains.kotlin.fir.resolve.toClassSymbol
import org.jetbrains.kotlin.fir.types.coneTypeOrNull

internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
  FirStatusTransformerExtension(session) {
  override fun FirDeclarationPredicateRegistrar.registerPredicates() {
    register(session.predicates.dependencyGraphPredicate)
  }

  override fun needTransformStatus(declaration: FirDeclaration): Boolean {
    // First check if this is an accessor in a dependency graph
    if (declaration !is FirCallableDeclaration) return false

    // If it's already an override, nothing needed here
    if (declaration.symbol.isOverride) return false

    // Only abstract callables
    if (
      declaration is FirSimpleFunction && declaration.hasBody ||
        (declaration as? FirProperty)?.getter?.hasBody == true
    ) {
      return false
    }

    if (declaration is FirSimpleFunction && declaration.valueParameters.isNotEmpty()) return false

    val containingClass = declaration.getContainingClassSymbol() ?: return false
    val isInGraph =
      session.predicateBasedProvider.matches(
        session.predicates.dependencyGraphPredicate,
        containingClass,
      )

    if (!isInGraph) return false

    // isAbstract doesn't work in interfaces
    if (containingClass.classKind?.isInterface != true && !declaration.isAbstract) return false

    // Exclude Provides/Binds
    val annotations = declaration.symbol.metroAnnotations(session)
    val isAccessor = !annotations.isProvides && !annotations.isBinds

    if (!isAccessor) return false

    return true
  }

  override fun transformStatus(
    status: FirDeclarationStatus,
    declaration: FirDeclaration,
  ): FirDeclarationStatus {
    if (declaration !is FirCallableDeclaration) return status
    val returnType = declaration.returnTypeRef.coneTypeOrNull ?: return status

    val containingClass = declaration.getContainingClassSymbol() ?: return status

    val name = declaration.symbol.callableId.callableName

    var needsOverride = false
    for (superType in containingClass.getSuperTypes(session, recursive = false)) {
      val classSymbol = superType.toClassSymbol(session) ?: continue
      if (
        !classSymbol.name.asString().startsWith(Symbols.StringNames.METRO_CONTRIBUTION_NAME_PREFIX)
      ) {
        continue
      }

      // We only want @ContributesTo types, which have supertypes
      val contributedInterface =
        classSymbol.resolvedSuperTypes.firstOrNull()?.toClassSymbol(session) ?: continue

      // Walk its direct callables. If any clash, mark needsOverride as true
      val matchingCallable =
        contributedInterface
          .callableDeclarations(session, includeSelf = true, includeAncestors = false)
          .firstOrNull { it.name == name && it.resolvedReturnType == returnType }

      if (matchingCallable != null) {
        needsOverride = true
        break
      }
    }

    if (!needsOverride) return status
    // TODO this doesn't seem to be quite enough, still fails in `FirOverrideChecker`. Maybe we need to add overridden symbols
    return status.copy(isOverride = true)
  }
}

```

I _think_ there is a bug in the compiler in FirOverrideChecker, as the issue seems to be that it checks `member.isOverride`, which only checks _rawStatus_ and not `resolvedStatus`. When I check in the debugger, it looks like rawStatus returns false (i.e. what was in source) but `resolvedStatus` returns true (what we changed). I'll file an issue on youtrack.

image

ZacSweers:
Took a shot at a PR https://github.com/JetBrains/kotlin/pull/5466

Downside though is that if that is the ultimate issue, it will be a minute until it's merged and you'll wanna use the workaround I mentioned above until then

ZacSweers:
Pushed up the transformer I started. Still not sure if this is the best approach considering the lookups it has to do, feel free to iterate on it!

kevinguitar:
@ZacSweers Thanks a lot! I'll keep an eye on your PR, and continue the effort here if it's necessary 👍 

The workaround works, I also figured out another workaround that works fine with the IDE. So I moved all the override-compatible accessors to another interface, and uses `asContribution` to grab them.
```kotlin
interface AppGraph {
  @Provides
  fun providesInt(): Int = 3
}

interface AppDependencies {
  val int: Int
}

interface ServiceProvider {
  val int: Int
}
```

I'm fine to close this PR to keep things tidy if you prefer, and we can reopen it if it's necessary.

ZacSweers:
We can keep this for now and revisit once the upstream change is available

ZacSweers:
Fix is merged and will be in 2.2.20-Beta2. Once I start a branch for those I'll re-target this one onto that

ZacSweers:
Retargeted to the 2.2.20 branch

ZacSweers:
To my great surprise it seems that the status transformer can receive java sources (??). Working around that for now but it's making me question if I understand what these transformers can process

kevinguitar:
Huh, curious how you found out, there are no java sources from the test case 🤔 

Btw, do we still need `FirAccessorOverrideStatusTransformer` aside from the fix you made to Kotlin?

ZacSweers:
The java sources are in some dagger interop tests that started failing

ZacSweers:
> Btw, do we still need FirAccessorOverrideStatusTransformer aside from the fix you made to Kotlin?

Yep we do. My fix on the kotlin side was to make this work

kevinguitar:
Ah got it 🤯

ZacSweers:
Merging this into the 2.2.20 branch 🥳

PR #670, opened 2025-07-04
Add diagnostic for scopes on binds annotations

### LINKED ISSUE #669: Report errors for scoped `@Binds` declarations
These are simply pipes and do not have their own scope

ZacSweers:
Resolves #669


kevinguitar:
Curious about the decision, does `@Binds` work in a different way than `@ContributesBinding` with scope?

ZacSweers:
Yes, ContributesBinding just tells the generator to generate a simple binds function. The source class itself may be annotated with a scope and that will still be respected in its own ConstructorInjected binding

PR #672, opened 2025-07-05
Allow empty multibinding for Dagger's `@Multibinds` interop

kevinguitar:
In metro, `Multibinds.allowEmpty` is default to false, but the behavior in Dagger is actually true, I found it during migration and needed to migrate to metro's annotation. In this PR, I made Dagger interop's value default to true to make the migration smoother.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
@@ -13,8 +13,8 @@ import dev.zacsweers.metro.compiler.ir.parameters.Parameters
 import dev.zacsweers.metro.compiler.isWordPrefixRegex
 import dev.zacsweers.metro.compiler.render
 import dev.zacsweers.metro.compiler.unsafeLazy
-import org.jetbrains.kotlin.backend.jvm.ir.parentClassId
 import java.util.TreeSet
+import org.jetbrains.kotlin.backend.jvm.ir.parentClassId
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrDeclaration
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationWithName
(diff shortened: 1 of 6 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
```diff
@@ -463,11 +465,20 @@ internal sealed interface Binding : BaseBinding<IrType, IrTypeKey, IrContextualT
         multibinds: IrAnnotation,
         contextualTypeKey: IrContextualTypeKey,
       ): Multibinding {
+        val isDaggerMultibinds =
+          multibinds.ir.symbol.owner.parentClassId?.asSingleFqName() ==
```
kevinguitar: Probably there is a better way to resolve classId or FqName

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
```diff
@@ -463,11 +465,20 @@ internal sealed interface Binding : BaseBinding<IrType, IrTypeKey, IrContextualT
...
+            FqName("dagger.multibindings.Multibinds")
```
kevinguitar: I saw `Symbols.DaggerSymbols.ClassIds`, let me know if it's better to move the declaration there

ZacSweers: yup let's define it there for reuse

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
ZacSweers: cleaned up in c9feb68d2847d06c12c30da31ea785d1236746b2

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/Binding.kt
```diff
@@ -463,11 +465,15 @@ internal sealed interface Binding : BaseBinding<IrType, IrTypeKey, IrContextualT
         multibinds: IrAnnotation,
         contextualTypeKey: IrContextualTypeKey,
       ): Multibinding {
+        // Retain Dagger's behavior in interop if using their annotation
+        val assumeAllowEmpty =
+          metroContext.options.enableDaggerRuntimeInterop &&
```
kevinguitar, line 470: Just for me to understand, is there a case that the project doesn't enable dagger interop, but we process dagger's `@Multibinds` here?

ZacSweers: It's possible yeah, you could just enable the annotations replacement

kevinguitar: ah right 👍

PR #673, opened 2025-07-06
Fix BaseDoubleCheck.isInitialized() check

kevinguitar:
Not sure if it affects anything, let's see from CI

### DIFF
--- a/runtime/src/commonMain/kotlin/dev/zacsweers/metro/internal/BaseDoubleCheck.kt
+++ b/runtime/src/commonMain/kotlin/dev/zacsweers/metro/internal/BaseDoubleCheck.kt
@@ -60,7 +60,7 @@ public abstract class BaseDoubleCheck<T>(provider: Provider<T>) :
       }
     }
 
-  override fun isInitialized(): Boolean = _value === UNINITIALIZED
+  override fun isInitialized(): Boolean = _value !== UNINITIALIZED
 
   override fun invoke(): T = value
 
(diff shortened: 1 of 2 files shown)


runtime/src/commonMain/kotlin/dev/zacsweers/metro/internal/BaseDoubleCheck.kt
```diff
@@ -60,7 +60,7 @@ public abstract class BaseDoubleCheck<T>(provider: Provider<T>) :
       }
     }
 
-  override fun isInitialized(): Boolean = _value === UNINITIALIZED
+  override fun isInitialized(): Boolean = _value !== UNINITIALIZED
```
ZacSweers, line 63: This is... weird as I'm pretty sure this was autoconverted from Java 🫣. There's some tests too

ZacSweers: Errr wait no this would be copied from the stdlib lazy, probably doesn't have any tests from dagger as this doesn't exist in its version. Would be good to add one as part of this fix

kevinguitar: Yeah Dagger lazy doesn't have isInitialized fun, do you expect more test cases here?

ZacSweers: Yeah if you don't mind adding one that would be good

kevinguitar: I don't mind, just need to know what scenario else to cover 🤔 I added one in this PR already, or do you expect tests for the BaseDoubleCheck level?

ZacSweers: Oops missed that on mobile 🫣

ZacSweers (approved):

PR #723, opened 2025-07-16
Add a new Metro option `warnOnInjectAnnotationPlacement`

kevinguitar:
Add a new Metro option `warnOnInjectAnnotationPlacement` to disable suggestion to lift @Inject to class when there is only one constructor. Also, adjusted how the existing check works, the warning applies to constructors with params as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog
 --------------
 
 - **Enhancement:** Chunk field initializers and constructor statements across multiple init functions to avoid `MethodTooLargeException` in large graphs.
+- **Enhancement:** Add a new Metro option `suggestClassInjectionIfNoParams` to disable suggestion to lift @Inject to class if the primary constructor is empty.
 
 0.5.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog
 
 - **Enhancement:** Chunk field initializers and constructor statements across multiple init functions to avoid `MethodTooLargeException` in large graphs.
 - **Enhancement:** Add a new Metro option `suggestClassInjectionIfNoParams` to disable suggestion to lift @Inject to class if the primary constructor is empty.
+- **Breaking change:** Rename the `generateHintProperties` Gradle DSL property to `generateContributionHints`.
 
 0.5.0
 -----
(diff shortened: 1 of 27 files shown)


kotlin-js-store/package-lock.json
ZacSweers, line 1: Can you revert these package-lock changes for now? I'm not sure what's up with these KGP tasks but they seem non-deterministic at the moment :/. Working on it separately

kevinguitar: Sure thing

gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroPluginExtension.kt
```diff
@@ -102,6 +102,12 @@ constructor(layout: ProjectLayout, objects: ObjectFactory, providers: ProviderFa
         providers.gradleProperty("metro.version.check").map { it.toBoolean() }.orElse(true)
       )
 
+  /**
+   * Enable/disable suggestion to lift @Inject to class if there is no params. Enabled by default.
+   */
+  public val suggestClassInjectionIfNoParams: Property<Boolean> =
```
ZacSweers: Let's call it `warnOnInjectAnnotationPlacement` and keep it generic, and update the other names accordingly. Let's also add a test for this to compiler-tests

compiler-tests/src/test/data/diagnostic/fir/inject/SuggestInjectClassOnSingleContructor.kt
```diff
@@ -0,0 +1,11 @@
...
+
+class SingleNonEmptyConstructor <!SUGGEST_CLASS_INJECTION!>@Inject<!> constructor(int: Int)
+
+class InjectPrimaryConstructor @Inject constructor(int: Int) {
+  constructor() : this(0)
+}
+
+class InjectSecondaryConstructor internal constructor(int: Int) {
+  @Inject constructor() : this(0)
+}
```
kevinguitar, line 11: These two cases are valid, we report warning only for single constructor now

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changelog
 --------------
 
 - **Enhancement:** Chunk field initializers and constructor statements across multiple init functions to avoid `MethodTooLargeException` in large graphs.
+- **Enhancement:** Add a new Metro option `warnOnInjectAnnotationPlacement` to disable suggestion to lift @Inject to class when there is only one constructor.
```
gildor: I think enhancement entry should also mention that new check works for classes with any amount of params, not only for classes with an empty constructor

kevinguitar: Adjusted a bit

PR #741, opened 2025-07-18
Fail tests eagerly when compiler plugins report errors

bnorm:
This avoids attempting to run box tests which are likely to fail at runtime if a compiler plugin reports errors in the backend. Tests should avoid the test runtime exception when possible and should only report the diagnostic difference.

### DIFF
--- a/compiler-tests/src/test/data/box/provides/SimpleFunctionProvider.kt
+++ b/compiler-tests/src/test/data/box/provides/SimpleFunctionProvider.kt
@@ -2,8 +2,8 @@
 interface ExampleGraph {
   val provider: Provider<String>
 
-//  @Provides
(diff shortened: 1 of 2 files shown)


compiler-tests/src/test/data/box/provides/SimpleFunctionProvider.kt
```diff
@@ -2,8 +2,8 @@
 interface ExampleGraph {
   val provider: Provider<String>
 
-  @Provides
-  fun provideValue(): String = "Hello, world!"
+//  @Provides
+//  fun provideValue(): String = "Hello, world!"
```
bnorm: Temporary to test GitHub build output.

PR #745, opened 2025-07-19
Remove references to @Multibinding

alexvanyo:
Just a mall fix to remove old references to `@Multibinding`.

I noticed this wasn't correct in the docs, and did a global search in the project for `@Multibinding`
to verify only correct usages were left, which caught one usage in a test as well.

No functional changes are in this PR.


ZacSweers (approved):
Thanks!

PR #746, opened 2025-07-19
Add failing tests for missing metadata binding container issues

JoelWilcox:


ZacSweers:
Folding this into #755

PR #755, opened 2025-07-20
Introduces binds mirrors and expect/actual IC tracking

### LINKED ISSUE #742: Metadata is not generated for graphs that only has `@Multibinds`
### Description

If a graph extension contains only `@Multibinds`, the root graph fails to compile because it can't find the metadata for the extension.

### Self-contained Reproducer

```kotlin
@ContributesGraphExtension(AppScope::class)
interface SomeGraphExtension {
    @Multibinds(allowEmpty = true)
    fun primeInitializable(): Set<Initializable>
}
```

### Metro version

0.5.1

### Context

_No response_

### LINKED ISSUE #743: No metadata found for no-op binding container
### Description

With Metro 0.5.1, I'm seeing the following error:
```
e: java.lang.IllegalStateException: No metadata found for com.package.SomeModule from another module. Did you run the Metro compiler plugin on this module?
        at dev.zacsweers.metro.compiler.ir.transformers.BindingContainerTransformer.loadExternalBindingContainer(BindingContainerTransformer.kt:646)
```

This gets triggered by what is essentially a no-op binding container. Our use-case for the no-op is that we have a custom code generator which references and generates code related to `SomeModule`, requiring it to always exist. Then we have two gradle modules that each define `SomeModule` in the same package, and some apps get the no-op while others get the functional implementation.

Would you be open to loosening the validation here? Maybe behind a flag if it's preferred to keep this an error by default

### Self-contained Reproducer

This is reproducible with an empty binding container in a separate module from the related graph declaration
```
@BindingContainer
interface SomeModule
```

### Metro version

0.5.1

### Context

_No response_

ZacSweers:
Resolves #743, resolves #742


ZacSweers:
Working on these two IC tests. Changing just the annotations on the two functions despite recording them to the lookup tracker isn't working for some reason 🤔

If that's gonna be an issue, we might need to generate some generated holder declaration with a hash of the annotation in the name

ZacSweers:
Looks like it's [this](https://youtrack.jetbrains.com/issue/KT-14440/Gradle-Annotation-list-changes-should-be-detected-and-used-in-incremental-compilation). Will look into generating mirror functions for these

PR #783, opened 2025-07-24
chore: Update copyright text to use current year

hossain-khan:
Not productive change, however when I noticed it, I thought why not! 🫣


ZacSweers:
We actually want to preserve the copyright to its inception year. Some projects do an updated inception -> current but I haven't historically dealt with that churn since the repo license and header in the files are enough

PR #797, opened 2025-07-25
Simplify shadowJar configuration

Goooler:


compiler/build.gradle.kts
```diff
@@ -63,6 +63,10 @@ val shadowJar =
         // or we'll get two artifacts that have the same classifier
         archiveClassifier.set("ignored")
       }
+
+      enableRelocation = true
+      relocationPrefix = "dev.zacsweers.metro.compiler.shaded"
```
ZacSweers, line 68: What determines which packages are relocated?

Goooler: The prefix will be added to all packages of depenencies.

ZacSweers: I do not want to shade all dependencies, only certain dependencies

PR #806, opened 2025-07-26
[For discussion] First Metrox lib, metrox-android

yschimke:
Add a metrox-android library that supports appComponentFactory automatically.

TODO

- [ ] Documentation
- [ ] Update samples - currently blocked on `includeBuild` issues
- [ ] Add tests

### DIFF
--- a/metrox/android/src/main/AndroidManifest.xml
+++ b/metrox/android/src/main/AndroidManifest.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+        xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application
+            android:appComponentFactory="dev.zacsweers.metro.sample.android.MetroAppComponentFactory"/>
+</manifest>
--- a/metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentFactory.kt
+++ b/metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentFactory.kt
@@ -1,0 +1,80 @@
+// Copyright (C) 2025 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metrox.android
+
+import android.app.Activity
+import android.app.Application
+import android.app.Service
+import android.content.BroadcastReceiver
+import android.content.ContentProvider
+import android.content.Intent
+import android.os.Build
+import androidx.annotation.Keep
+import androidx.annotation.RequiresApi
+import androidx.core.app.AppComponentFactory
+import dev.zacsweers.metro.Provider
+import kotlin.reflect.KClass
+
+/**
+ * An [AppComponentFactory] that uses Metro for constructor injection of Activities.
+ *
+ * If you have minSdk < 28, you can fall back to using member injection on Activities or (better)
+ * use an architecture that abstracts the Android framework components away.
+ */
+@RequiresApi(Build.VERSION_CODES.P)
+@Keep
+class MetroAppComponentFactory : AppComponentFactory() {
+
+  private inline fun <reified T : Any> getInstance(
+    cl: ClassLoader,
+    className: String,
+    providers: Map<KClass<out T>, Provider<T>>,
+  ): T? {
+    val clazz = Class.forName(className, false, cl).asSubclass(T::class.java)
+    val modelProvider = providers[clazz.kotlin] ?: return null
+    return modelProvider()
+  }
+
+  override fun instantiateActivityCompat(
+    cl: ClassLoader,
+    className: String,
+    intent: Intent?,
+  ): Activity {
+    return getInstance(cl, className, metroAndroidAppGraph.activityProviders)
+      ?: super.instantiateActivityCompat(cl, className, intent)
+  }
+
+  override fun instantiateApplicationCompat(cl: ClassLoader, className: String): Application {
+    val app = super.instantiateApplicationCompat(cl, className)
+    metroAndroidAppGraph = (app as MetroApplication).appGraph
+    return app
+  }
+
+  override fun instantiateProviderCompat(cl: ClassLoader, className: String): ContentProvider {
+    return getInstance(cl, className, metroAndroidAppGraph.providerProviders)
+      ?: super.instantiateProviderCompat(cl, className)
+  }
+
+  override fun instantiateReceiverCompat(
+    cl: ClassLoader,
+    className: String,
+    intent: Intent?,
+  ): BroadcastReceiver {
+    return getInstance(cl, className, metroAndroidAppGraph.receiverProviders)
+      ?: super.instantiateReceiverCompat(cl, className, intent)
+  }
+
+  override fun instantiateServiceCompat(
+    cl: ClassLoader,
+    className: String,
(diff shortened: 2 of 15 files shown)


metrox/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+        xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application
+            android:appComponentFactory="dev.zacsweers.metro.sample.android.MetroAppComponentFactory"/>
```
SteinerOk: Incorrect, must be `dev.zacsweers.metrox.android.MetroAppComponentFactory`

SteinerOk:
> Discuss the pre-API 28 approach
I think we should have a solution for API 23+, since many applications use it.

+
I think it would be good to think about Fragments and ViewModels, WorkManager

yschimke:
Yep, discussed in the Slack channel. I think those can be separate modules.

Metrox-work with dependency on work manager

Metrox-viewmodel with CMP view model support

SteinerOk:
@yschimke Yes, I agree, separate modules are good

It would be great to have support for extended graphs (AppGraph -> AuthedGraph/UnauthedGraph, as sample)

hossain-khan:
Great initiative, as much as I would love to have something like this, I feel like Metro itself has massive domain to cover being a multiplatform framework. Adding `metrox` at this point will add overhead to maintaining it as it moves forward fast 😅 

Just my personal feeling, I don't have any specifics to back this up! 🙃

yschimke:
Yep, but it also makes the case for providing it. If it's so complicated, brittle or subject to churn that each metro release risks breaking it, then what exactly are we expecting developers to do?

SteinerOk:
@yschimke @ZacSweers I think a good solution, focused on long-term development, would be to create a separate "GitHub organization", for example `metro-di` look like `sqldelight` has, with repositories in it `metro`,` metrox-android`, `metrox-ios`, and so on.
And the repository ` metrox-android` will already contain separate modules for fragments, appcomponents(-preapi28)

yschimke:
Sure, there is no rush with this. If you or one of the other maintainers is planning to set those orgs up, I can wait and contribute a PR there.

Alternatively, if you don't think there is enough value in reusable libraries owned by the maintainers, I can close the PR and we just leave it as samples.

dewantawsif:
As Dagger2 has Hilt for simplifying development on Android it'll be nice to also have `metrox` family to simplify development on Android, JVM etc.

ZacSweers (commented):
I think I'm game to merge this in. One thing I'm a bit stuck on though is `MetroAndroidAppGraph`. Does this _need_ to be a dependency graph or can we just make this an interface other folks can extend in their app graphs? Especially now that graph extensions are no longer detached

ZacSweers:
> Discuss the pre-API 28 approach

I think your doc basically saying pre-28 requires member injection is enough for me

yschimke:
I'll rebase the change and take a look at your suggestion.

ZacSweers:
Made a few changes. One thing I'm stumped on right now is why gradle's dependency substitution isn't working for that artifact 🤔

metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentProviders.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @see MetroAppComponentFactory
+ * @see MetroApplication
+ */
+interface MetroAppComponentProviders {
+  /**
+   * A multibinding map of [Activity] classes to their providers accessible for
+   * [MetroAppComponentFactory].
+   */
+  @Multibinds val activityProviders: Map<KClass<out Activity>, Provider<Activity>>
```
hrach: Not sure if for other activities, but definitely other multibinds should be `allowEmpty = true` as they may not be used at all.

metrox-android/src/main/kotlin/dev/zacsweers/metrox/android/MetroApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+   * A [MetroAppComponentProviders] instance to access
+   *
+   * ```kotlin
+   * class MetroApp : Application(), MetroApplication {
+   *   /** Holder reference for the app graph for [MetroAppComponentFactory]. */
+   *   val appGraph by lazy { createGraph<AppGraph>() }
+   *
+   *   override val appComponentProviders: MetroAppComponentProviders
+   *     get() = appGraph
+   * }
```
hrach, line 23: AFAIK, this got out of sync after refactoring. Now the AppGraph does not implement MetroAppComponentProviders.

metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentProviders.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * An interface of common Map multibinding providers for Android "app components" like Activities,
+ * Services, etc. This is intended for use with [MetroAppComponentFactory] to allow for easy
+ * constructor injection of these types with minSdk 28+.
+ *
+ * @see MetroAppComponentFactory
+ * @see MetroApplication
+ */
+interface MetroAppComponentProviders {
```
hrach: I'm not sure if I missed something, but after reworking my app to this latest state of the PR, it does not compile: 

```
Cannot find an @Inject constructor or @Provides-annotated function/property for: MetroAppComponentProviders
```

I.e., who implements this? As a dependency graph no longer inherits it, there is no impl, is it?

metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentFactory.kt
```diff
@@ -17,6 +20,7 @@ import kotlin.reflect.KClass
  * If you have minSdk < 28, you can fall back to using member injection on Activities or (better)
  * use an architecture that abstracts the Android framework components away.
  */
+@RequiresApi(28)
 @Keep
 class MetroAppComponentFactory : AppComponentFactory() {
```
hrach: Shouldn't it inherit `androidx.core.app.CoreComponentFactory`?

ZacSweers:
@hrach appreciate the comments but don't need the extra review, thanks!

metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentProviders.kt
ZacSweers: The graph implements it, that is defined in the consuming app

metrox/android/src/main/kotlin/dev/zacsweers/metrox/android/MetroAppComponentFactory.kt
ZacSweers: That's not a public API

ZacSweers:
Abandoning setting up a unit test for this for now due to https://github.com/robolectric/robolectric/pull/10833, and I'm not willing to go through the trouble of setting up emulator/device instrumentation tests for this project's CI

PR #809, opened 2025-07-26
Unify FIR and IR diagnostic tests

bnorm:
With `AbstractPhasedJvmDiagnosticLightTreeTest` as the base class, there's no reason to have 2 diagnostic test directories anymore. Defaults to expect things to fail in the frontend, but can override to specify FIR2IR in the individual tests.


ZacSweers (approved):
Nice

Issue #664, opened 2025-07-04
enableScopedInjectClassHints not working for @Inject constructor


blackmailcz:
### Description

Hello,

refering to #377 , where the flag `enableScopedInjectClassHints` does not apply correctly for `@Inject constructor(...)` syntax.

### Self-contained Reproducer

```kotlin
import dev.zacsweers.metro.AppScope
import dev.zacsweers.metro.ContributesGraphExtension
import dev.zacsweers.metro.DependencyGraph
import dev.zacsweers.metro.Inject
import dev.zacsweers.metro.SingleIn

sealed interface LoggedInScope

@SingleIn(AppScope::class) class Dependency @Inject constructor()
@Inject @SingleIn(LoggedInScope::class) class ChildDependency(val dep: Dependency)

@DependencyGraph(scope = AppScope::class, isExtendable = true)
interface ExampleGraph {
    // Works if added explicitly like this
    // val dependency: Dependency
}

@ContributesGraphExtension(LoggedInScope::class)
interface LoggedInGraph {
    val childDependency: ChildDependency

    @ContributesGraphExtension.Factory(AppScope::class)
    interface Factory {
        fun createLoggedInGraph(): LoggedInGraph
    }
}
```

### Metro version

0.4.0

### Context

The problem seems to be in `IrContributionVisitor:55`, where the condition explicitely checks only for `declaration` annotation. It should check if it has a constructor with `@Inject` as well

Problematic snippet:

```kotlin
    if (
      options.enableScopedInjectClassHints &&
        declaration.isAnnotatedWithAny(symbols.classIds.injectAnnotations) &&
        !declaration.isAnnotatedWithAny(symbols.classIds.allContributesAnnotations)
    ) {
      data.addScopedInject(declaration)
    }
```

PR #822, opened 2025-07-28
Failing test: Parent scoped class is unused in parent graph

kevinguitar:
Regression of https://github.com/ZacSweers/metro/issues/664

Test fails with
```
Multiple Failures (2 failures)
	org.opentest4j.AssertionFailedError: Actual data differs from file content: ParentScopedClassUnusedInParent.kt
	java.lang.IllegalStateException: //private/var/folders/pt/7n9vn6c131s2yz1bww3j3mf80000gp/T/dev.zacsweers.metro.compiler.BoxTestGenerated$DependencygraphtestParentScopedClassUnusedInParent/kotlin-sources/main/ParentScopedClassUnusedInParent.kt:(115,126): error: [Metro/IncompatiblyScopedBindings] ParentGraph.$$ContributedChildGraph (scopes '@SingleIn(Unit::class)') may not reference bindings from different scopes:
    Foo (scoped to '@SingleIn(AppScope::class)')
    Foo is requested at
        [ParentGraph.$$ContributedChildGraph] ChildGraph#foo


(Hint)
$$ContributedChildGraph is contributed by 'ChildGraph' to 'ParentGraph'.

(Hint)
It appears that extended parent graph 'ParentGraph' does declare the '@SingleIn(AppScope::class)' scope but doesn't use 'Foo' directly.
To work around this, consider declaring an accessor for 'Foo' in that graph (i.e. `val foo: Foo`) or enabling the `enableScopedInjectClassHints` option.
See https://github.com/ZacSweers/metro/issues/377 for more details.
```

### DIFF
--- a/compiler-tests/src/test/data/box/dependencygraph/ParentScopedClassUnusedInParent.kt
+++ b/compiler-tests/src/test/data/box/dependencygraph/ParentScopedClassUnusedInParent.kt
@@ -1,3 +1,4 @@
+// Regression test for https://github.com/ZacSweers/metro/issues/664
 @DependencyGraph(AppScope::class, isExtendable = true)
 interface ParentGraph
 


kevinguitar:
Ah I didn't realize that `enableScopedInjectClassHints` is disabled by default in the box test. Added `ENABLE_SCOPED_INJECT_CLASS_HINTS` directive works. Closing, will open a new one once I manage to reproduce our case

PR #834, opened 2025-07-28
Add a failing test for a child graph accessing an empty multibinding from a binding container contributed to its parent

JoelWilcox:


ZacSweers:
Fixed on main and cherrypicked this test with it!

PR #840, opened 2025-07-30
CI Workflow: Add GitHub Actions workflow for deploying Metro docs site to GitHub Pages

hossain-khan:
## Summary
Added dedicated workflow to continuously deploy latest docs changes to the GitHub pages site.


.github/workflows/docs-site.yml
```diff
@@ -0,0 +1,94 @@
...
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  # Metro docs site build job
+  docs-site:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4 # Or, actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
hossain-khan: Not sure why specific hash was used for this action! Should I match it?

.github/workflows/docs-site.yml
```diff
@@ -0,0 +1,94 @@
+# Generates and deploys static site to GitHub Pages
```
hossain-khan, line 1: 🤔 Thinking if I am forgetting to update any existing code/script that may be related to it.

hossain-khan (commented):
@ZacSweers This should take care of the docs site deployment automatically.

I couldn't do dry-run! however, once merged it should kick off the job and I can debug any issue from the log, except any permission issue that workflow needs.

.github/workflows/docs-site.yml
```diff
@@ -0,0 +1,94 @@
...
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
+
+          # Don't reuse cache entries from any other Job.
+          gradle-home-cache-strict-match: true
+
+          cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
+
+      - name: Generate API docs
+        # NOTE: The `dokkaPublications.html` task is pre-configured to generate HTML in `/docs/api`.
+        run: ./gradlew :dokkaGenerate --rerun-tasks --no-build-cache
```
hossain-khan: Also this command keeps failing on me in the codespaces, likely related to low memory? not sure.

```
> Task :runtime:commonizeNativeDistribution FAILED

* What went wrong:
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

However, this did work:

```sh
./gradlew :dokkaGenerate -x :runtime:commonizeNativeDistribution
```

ZacSweers: yah that sounds like low memory

.github/workflows/docs-site.yml
ZacSweers: renovate will use specific hashes by default for idempotence

ZacSweers:
```suggestion
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```

.github/workflows/docs-site.yml
```diff
@@ -0,0 +1,94 @@
...
+          GRADLE_OPTS: -Xmx4g # Dokka can be memory intensive, so we increase the heap size
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Install MkDocs dependencies
+        run: |
+          pip install mkdocs mkdocs-material mdx_truly_sane_lists "mkdocs-material[imaging]"
```
ZacSweers: Any reason not to use the requirements.txt file in the repo?

hossain-khan: That's a good idea. I'll update.

hossain-khan:
> Can we make this workflow reuse the deploy_website.sh script at all to de-dupe?

I think the reason that script exist is to clone repo into isolated directory, and then apply these changes. 

However, on CI it's already isolated per job. So we can apply these changes directly.

I think a good move would be to refactor the script into few pieces so that those can be used in both CI and in the deploy_website.sh

I'll take a look.

ZacSweers:
First run is here! https://github.com/ZacSweers/metro/actions/runs/16613166178

ZacSweers:
Welp it failed

hossain-khan:
Not bad! Just needs the new dep in requirements!

```
Run mkdocs build
ERROR   -  Required dependencies of "social" plugin not found:
- ModuleNotFoundError("No module named 'cairosvg'")
- ModuleNotFoundError("No module named 'PIL'")

--> Install with: pip install "mkdocs-material[imaging]"
```

PR #854, opened 2025-07-30
Added test/docs for providing from a BindingContainer constructor property parameter

jonapoul:
I starting putting this together because I hadn't been able to get constructor injection to work from a `BindingContainer` when passed into a graph via the factory. I was digging around in the IR processing logic to see how it would work, then I found out that it does already work - I had just been approaching it in the wrong way!

My initial attempt involved:
```kotlin
@BindingContainer
class MyContainer(
  @param:Provides val paramA String,
  @Provides val paramB: Int,
  val paramC: Double
) {
  @Provides fun c(): Double = paramC
}
```
A and B don't work, and C does but it's a bit more verbose than I'd like. 

But as part of running the debugger I found out that metro already does support it, using `@get:Provides` instead! So here's a test to confirm it, plus some additions to the docs to write it down a bit more explicitly.


ZacSweers:
I'm open to supporting this, the behavior you're currently seeing is more coincidence. Let me fix this to support without requiring the `@get:`

jonapoul:
> the behavior you're currently seeing is more coincidence

I had a hunch that might be the case!

ZacSweers:
I'm actually going to disallow this behavior after further consideration. There are too many unknowns to handle this properly

- What if multiple constructors are defined? How can we know which ones to read from?
- The compiler does not currently propagate annotations onto both parameters _and_ properties. It is introducing this behavior in a future version, but it seems to be an IR-intrinsic. Thus, we'd have to read parameters in the constructor first and then try to fuzzy match them to properties that initialize to them. I'm not sure how to do this reliably right now.

PR #867, opened 2025-08-01
Fix handling repeated annotations from external modules

kevinguitar:
I noticed a bug with multiple replacers from the same class, there is a precondition - replacers need to live in a different module from the graph.

The test failed with:
```
error: [Metro/DuplicateBinding] Duplicate binding for Bar
├─ Binding 1: Contributed by 'FakeImpl' at an unknown source location (likely a separate compilation).
├─ Binding 2: Contributed by 'RealImpl' at an unknown source location (likely a separate compilation).
```

PR #868, opened 2025-08-01
IC test: Multiple binding replacements are respected when adding new binding contribution

kevinguitar:
Few folks from my team including myself, started to face an IC issue related to multiple replacements from a single class. It's a bit different than https://github.com/ZacSweers/metro/pull/867, in this scenario, the replacers live in the same module as the graph.

The mutator could be something else, but adding a new binding contribution seems the most reliable way to reproduce it.

The test fails with:
```
e: file:///Users/kevin/Documents/metro/gradle-plugin/build/functionalTest/-1f10b796-46/src/main/kotlin/test/AppGraph.kt:7:11 [Metro/DuplicateBinding] Duplicate binding for test.Bar
├─ Binding 1: Contributed by 'test.FakeImpl' at an unknown source location (likely a separate compilation).
├─ Binding 2: Contributed by 'test.RealImpl' at an unknown source location (likely a separate compilation).
```


ZacSweers:
My hope is that this is fixed along the way by #867 🤞

ZacSweers:
Welp, let me dig more

ZacSweers:
Figured it out. In the second recompilation, the FIR supertype generator gets a pre-compiled contributed type from the previous compilation (even though it's the same source set). So I just needed to tweak it to always check for repeated annotation containers.

PR #876, opened 2025-08-02
Do not apply the shadow plugin, use the task directly instead

martinbonnin:
See https://github.com/GradleUp/shadow/issues/1540

Instead of having `shadow` configure the `shadowJar` task, do it manually. This gives us more control over the output.

Especially, we can now replace the outgoing jar completely so that there is only one "runtime" outgoing variant and we don't have to debug Gradle selection algorithms.

Note: I believe the underlying issue is https://youtrack.jetbrains.com/issue/KT-53477. Once this is fixed, there is probably no need for shadowing anymore?


samples/integration-tests/build.gradle.kts
```diff
@@ -99,9 +100,4 @@ tasks.withType<Test>().configureEach {
   maxParallelForks = Runtime.getRuntime().availableProcessors() * 2
 }
 
-// See https://github.com/GradleUp/shadow/issues/1540
-configurations.configureEach {
-  if (name.startsWith("kotlinCompilerPluginClasspath")) {
-    attributes { attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.SHADOWED)) }
-  }
-}
+tasks.withType<KotlinJsTest>().configureEach { enabled = false }
```
martinbonnin: That is unrelated but fixed `:integration-tests:build` (see https://github.com/ZacSweers/metro/actions/runs/16689036869/job/47243849606)

I _think_ there were no JS tests in that module but feel free to revert if needed.

compiler/build.gradle.kts
```diff
@@ -53,49 +58,64 @@ tasks.test {
...
+val embedded = configurations.resolvable("embedded")
 
-val shadowJar =
-  tasks.shadowJar.apply {
-    configure {
-      if (isPublishing) {
-        // Since we change the classifier of the shadowJar we need to disable the default jar task
-        // or we'll get two artifacts that have the same classifier
-        archiveClassifier.set("ignored")
-      }
```
martinbonnin, line 65: I'm not sure why this was needed. If I'm missing something, feel free to revert.

ZacSweers: Probably copied over from another project for bespoke reasons

PR #883, opened 2025-08-04
WIP dagger interop fixes

ZacSweers:
Iterating on a few things while working through a dagger/anvil-heavy codebase


ZacSweers:
Some of these changes need some tests in-repo but will add those later

PR #762, opened 2025-07-21
Add an assisted injection sample to compose-navigation-app

yschimke:
This commit introduces an `AssistedCounterViewModel` that uses assisted injection for its `initialValue`.

Key changes:
- Created `AssistedCounterViewModel` with an `@AssistedFactory` for creating instances with a dynamic `initialValue`.
- Added `AssistedCounterScreen` to display and interact with the `AssistedCounterViewModel`.
- Updated `MainActivity` to include a navigation route to `AssistedCounterScreen`.
- Modified `ViewModelGraph` to include the `AssistedCounterViewModel.Factory`.
- Refactored `MetroViewModelFactory` to expose a `viewModelGraph()` function for creating `ViewModelGraph` instances.
- Introduced a new `metroViewModel` composable function that takes a factory lambda for creating ViewModels, allowing for assisted injection.

PR #886, opened 2025-08-05
Added a sample project for multi-module ViewModel injection

jonapoul:
I've been playing around with introducing this into my playground app with a bit of difficulty, so I had a crack at putting together a PoC multi-module sample app with bindings applied to VM constructors and/or assisted factories as necessary.

It's very barebones and could hit runtime errors if you bind the wrong key types, but it does work! There are probably more effective methods of doing these - any feedback welcome of course.

Follows on from the sample app in https://github.com/ZacSweers/metro/pull/762 and the discussion in https://github.com/ZacSweers/metro/discussions/750 - but specifically trying to solve the issues brought in by a multi-module setup instead of single-module.


gradle/libs.versions.toml
```diff
@@ -102,11 +105,14 @@ compose-runtime = { module = "org.jetbrains.compose.runtime:runtime", version.re
 compose-ui = { module = "org.jetbrains.compose.ui:ui", version.ref = "compose-jb" }
 
 coroutines = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinx-coroutines" }
+coroutines-swing = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-swing", version.ref = "kotlinx-coroutines" }
 coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "kotlinx-coroutines" }
 
 dagger-runtime = { module = "com.google.dagger:dagger", version.ref = "dagger" }
 dagger-compiler = { module = "com.google.dagger:dagger-compiler", version.ref = "dagger" }
 
+jetbrains-navigation-desktop = "org.jetbrains.androidx.navigation:navigation-compose:2.9.0-beta04"
```
jonapoul: The androidx version of compose navigation has released a JVM artifact, but it only includes stubs which fail at runtime (for whatever reason). This one works but [only has pre-releases available](https://mvnrepository.com/artifact/org.jetbrains.androidx.navigation/navigation-compose)

devkbersztolc:
Hi,
Thanks for sharing this.
I've been struggling with multiplatform's viewModel injection in multimodule project, where I want to have viewModel-metro stuff in one of the library submodule and use it from the app's features. So far without luck (viewModel graph doesn't see viewModels). I'll definitely take a look at your implementation!

devkbersztolc:
@jonapoul 
Have you tried to make it work with ViewModels marked as internal?

jonapoul:
@devkbersztolc You'll need to set the parent graph to internal I think, but otherwise it should work. Did you try it?

ZacSweers:
I have seen this btw, just heads down on some other work at the moment!

jonapoul:
No rush!

hrach:
Thank you for this! I finally successfully migrated from Koin to Metro. (multimodule, 100+ viewmodels) ❤️ 🎉 

A few implementation notes:
- I introduced only assisted factories to the codebase, as the "viewModel" impl is not compile-time safe, and it could be an error-prone touch point - e.g., when a ViewModel is migrated to an assisted variant and the usage is kept using the non-assisted version.
- The implementation I used is using a "scope" to avoid defining two types (VM factory and VM): `metroViewModel { with<MyViewModel.Factory>().create(...) }`

<details>

```kotlin
@Composable
inline fun <reified VM : ViewModel> metroViewModel(
	viewModelStoreOwner: ViewModelStoreOwner = checkNotNull(LocalViewModelStoreOwner.current) {
		"No ViewModelStoreOwner was provided via LocalViewModelStoreOwner"
	},
	key: String? = null,
	noinline factory: ViewModelFactoryGraphScope.() -> VM,
): VM =
	viewModel(
		viewModelStoreOwner = viewModelStoreOwner,
		key = key,
		factory = (LocalActivity.current as HasDefaultViewModelProviderFactory).defaultViewModelProviderFactory,
		extras = when (viewModelStoreOwner) {
			is HasDefaultViewModelProviderFactory -> viewModelStoreOwner.defaultViewModelCreationExtras
			else -> CreationExtras.Empty
		}.withCreationCallback(factory),
	)

@PublishedApi
internal fun CreationExtras.withCreationCallback(callback: ViewModelFactoryGraphScope.() -> ViewModel): CreationExtras =
	MutableCreationExtras(this).apply {
		this[MetroViewModelFactory.ASSISTED_FACTORY_CALLBACK_KEY] = callback
	}

class ViewModelFactoryGraphScope(
	@PublishedApi
	internal val graph: ViewModelGraph,
) {
	inline fun <reified VMF : ViewModelFactory> with(): VMF =
		graph.assistedFactoryProviders[VMF::class]!!.invoke() as VMF
}

@Inject
class MetroViewModelFactory(
	private val graphProvider: ViewModelGraphProvider,
) : ViewModelProvider.Factory {
	companion object {
		val ASSISTED_FACTORY_CALLBACK_KEY: CreationExtras.Key<ViewModelFactoryGraphScope.() -> ViewModel> =
			object : CreationExtras.Key<ViewModelFactoryGraphScope.() -> ViewModel> {}
	}

	override fun <T : ViewModel> create(modelClass: Class<T>, extras: CreationExtras): T {
		val viewModelGraph = graphProvider.buildViewModelGraph(extras)
		val viewModelFactoryGraphScope = ViewModelFactoryGraphScope(viewModelGraph)
		val viewModelFactoryCallback = requireNotNull(extras[ASSISTED_FACTORY_CALLBACK_KEY]) {
			"Missing assisted factory callback. Use metroViewModel {}."
		}

		@Suppress("UNCHECKED_CAST")
		val viewModel = viewModelFactoryCallback.invoke(viewModelFactoryGraphScope) as T

		return viewModel
	}
}

```

</details>

JohannesPtaszyk:
Thank you already for the PR! 
I just tried porting your ideas to my project, but sadly not successful on first try, but on second.

Personally would love to split all classes into their own files instead of having multiple things in one file, so people reading through this can get a quicker and easier overview :)


And one small thing: I think it would be a huge advantage if your sample would have additional targets, like Android, so we can have a complete KMP + Compose project sample :)

ZacSweers:
@JohannesPtaszyk let's move that to a suggestion in discussions to avoid scope creep here 👍. It would be helpful if you could be more detailed there too on what you're looking for ("sadly not successful" isn't super actionable, and there are already other KMP+compose samples in this repo so are you asking for specifically one with KMP viewmodels?)

JohannesPtaszyk:
Will Open an issue with what I would expect as " full example" ☺️

ZacSweers:
Please use the discussions feature, not the issue tracker 👌

jonapoul:
Is this still of any interest? I'll rebase/fix if so

ZacSweers:
yes, sorry for the delays as I've been busy this week with travel. Don't worry about rebasing, I'm going to take it over and do some cleaning up. One thing I did wonder though is if there's any reason not to just expand the existing compose-navigation-app sample to just support multi-module? As it feels like this has a lot of overlap

jonapoul:
> Don't worry about rebasing, I'm going to take it over and do some cleaning up.

Cool, let me know if you need anything else from me :+1:  

> One thing I did wonder though is if there's any reason not to just expand the existing compose-navigation-app sample to just support multi-module? As it feels like this has a lot of overlap

Personally I see value in both - the existing one is great for getting a basic handle on how things can work on an introductory level and this is more of a scaled-up "realistic" implementation. Of course I can see that keeping both of them up to date adds maintenance effort so ultimately it's up to you, but I think people will find both useful in different ways

ZacSweers:
I'd argue the existing one isn't simple either and doing this kind of architectural plumbing is already a bit hands-on, so I'm leaning toward merging

ZacSweers:
@jonapoul did you still want to finish this?

jonapoul:
Can do - are you not taking it over?

ZacSweers:
I can! Just wanted to double-check before I do

jonapoul:
Be my guest!

ZacSweers:
Merged them and tidied things up a bit. Thanks! CC @yschimke

PR #898, opened 2025-08-08
fixed: re-enable metro docs site search

### LINKED ISSUE #897: Metro docs site's search feature got disabled, re-enable it.
As reported here in [ASG](https://androidstudygroup.slack.com/archives/C08CVCEFMLL/p1754656069584459?thread_ts=1754654360.820699&cid=C08CVCEFMLL), search feature is no longer visible.

### Root cause: 
**This is a common MkDocs "gotcha" - adding any plugin disables all default plugins unless explicitly included.**

So, search was disabled when the social card plugin was enabled in following PR
* https://github.com/ZacSweers/metro/pull/817

## Fix
Explicitly add search plugin in the mkdocs config.

```yml
plugins:
  # Social plugin for delightful card previews on social media
  # https://squidfunk.github.io/mkdocs-material/plugins/social/
  - social
  - search
```

👀 I will work on it tonight!

<details><summary>Additional background on the issue</summary>

```
  Search is enabled by default in MkDocs Material, but there's an important caveat:

  - The search plugin is enabled by default when no other plugins are specified
  - However, it must be explicitly re-added to mkdocs.yml when other plugins are used

  In Metro's case:
  1. Before July 27, 2025: No plugins section existed, so search was enabled by default
  2. After July 27, 2025: When the social plugin was added, search would have been automatically disabled unless
  explicitly re-added to the plugins list

  So the search functionality was likely working by default until the social plugin was added, at which point it
  would have been disabled.
```

</details>

hossain-khan:
It was unintentionally disabled. See https://github.com/ZacSweers/metro/issues/897 for details.
- [x] Validated search working by running it locally.

Fixes #897

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -45,6 +45,7 @@ plugins:
   # Social plugin for delightful card previews on social media
   # https://squidfunk.github.io/mkdocs-material/plugins/social/
   - social
+  - search
 
 #extra_css:
 #  - 'css/app.css'


ZacSweers (approved):
Thanks!

PR #925, opened 2025-08-15
Delete interop for non-existent MergeSubcomponent.Factory

Egorand:
https://github.com/square/anvil/blob/main/annotations/src/main/java/com/squareup/anvil/annotations/MergeSubcomponent.kt


ZacSweers (approved):
D'oh

PR #936, opened 2025-08-15
Fix Kotlin internal error overriding Metro error when there's a missing factory for a Java `@Inject` class

### LINKED ISSUE #935: 'Unknown file' Error message when no factory is found for a Java class with an `@Inject` constructor
### Description

When a Java class with an `@Inject` constructor doesn't have a factory created, we get an internal Kotlin error instead of a Metro-defined error pointing to the problematic class:
```
> Task :my:module:compileDebugKotlin FAILED
e: kotlin.NotImplementedError: An operation is not implemented: Unknown file
        at org.jetbrains.kotlin.ir.util.IrUtilsKt.getFile(IrUtils.kt:522)
        at org.jetbrains.kotlin.ir.KtDiagnosticReporterWithImplicitIrBasedContext.at(KtDiagnosticReporterWithImplicitIrBasedContext.kt:46)
        at dev.zacsweers.metro.compiler.ir.transformers.InjectConstructorTransformer.getOrGenerateFactory(InjectConstructorTransformer.kt:144)
```

The use-case I encountered this in is migrating from Dagger-Anvil and removing Dagger from a module that still had some `@Inject` classes written in Java.

### Self-contained Reproducer

Java:
```
public class ExampleClass {
              @Inject public ExampleClass() {
            
              }
            }
```

Kotlin:
```
            @DependencyGraph
            interface ExampleGraph {
              val exampleClass: ExampleClass
            }
```

Will put up a PR with test+fix shortly

### Metro version

0.6.1-SNAPSHOT

### Context

_No response_

JoelWilcox:
- Fixes #935

### DIFF
--- a/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/InjectConstructorTransformerTest.kt
+++ b/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/InjectConstructorTransformerTest.kt
@@ -4,10 +4,12 @@ package dev.zacsweers.metro.compiler.transformers
 
 import com.google.common.truth.Truth.assertThat
 import com.tschuchort.compiletesting.KotlinCompilation
+import com.tschuchort.compiletesting.KotlinCompilation.ExitCode
 import dev.zacsweers.metro.compiler.ExampleClass
 import dev.zacsweers.metro.compiler.ExampleGraph
 import dev.zacsweers.metro.compiler.MetroCompilerTest
 import dev.zacsweers.metro.compiler.assertCallableFactory
+import dev.zacsweers.metro.compiler.assertContainsAll
 import dev.zacsweers.metro.compiler.assertDiagnostics
 import dev.zacsweers.metro.compiler.assertNoArgCallableFactory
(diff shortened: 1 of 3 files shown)


compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/InjectConstructorTransformerTest.kt
```diff
@@ -359,4 +361,43 @@ class InjectConstructorTransformerTest : MetroCompilerTest() {
       )
     }
   }
+
+  // https://github.com/ZacSweers/metro/issues/935
+  // Covers a use-case where we would previously get a Kotlin internal error that read:
+  // e: kotlin.NotImplementedError: An operation is not implemented: Unknown file
+  @Test
+  fun `when no factory is found for a Java class, the exception should still specify it in the error message`() {
```
JoelWilcox, line 369: Wasn't able to use the compiler test framework for this one since diagnostics only cover FIR and box tests don't seem to have the ability to access/assert against error logs yet.

ZacSweers (approved):

Issue #983, opened 2025-08-22
[Dagger Interop] Graph extensions can't access providers from nested/included modules


erawhctim:
### Description

Ran into this issue with Dagger interop/annotations enabled; not sure if I'm hitting an actual bug, or just (incorrectly) expecting some prior Dagger behavior to work in Metro.

Assume a root graph like this, where a module is included at runtime in the graph factory and _that_ module contains another module via its `@Module(includes)` array:
```
@DependencyGraph
interface AppGraph {
  fun inject(app: ExampleApp)

  val extensionGraphFactory: ExtensionGraph.Factory

  @DependencyGraph.Factory
  interface Factory {
    fun create(@Includes appModule: ApplicationModule): AppGraph
  }
}

@Module(includes = [AppInfoModule::class])
class ApplicationModule(val application: Application) {
  @Provides fun app(): Application = application
}

@Module
class AppInfoModule {
  @Provides
  fun appInfo(application: Application): AppInfo = AppInfo(
    application.packageManager.getPackageInfo(application.packageName, 0).versionName
      ?: "unknown"
  )
}

data class AppInfo(val version: String)
```

If I define a graph extension like this, member injection via the extended graph can't access the provider from `AppInfoModule`
```
@GraphExtension(ExtensionScope::class)
interface ExtensionGraph {
  fun inject(a: ExtensionActivity)

  @GraphExtension.Factory
  interface Factory {
    fun create(@Includes module: ExtensionActivityModule): ExtensionGraph
  }
}

abstract class ExtensionScope private constructor()

@Module
class ExtensionActivityModule(val activity: ExtensionActivity) {
  @Provides fun provideExtAct(): ExtensionActivity = activity
}
```

Injecting an `AppInfo` instance works correctly from the root graph, but not when injected via the graph extension:
```
class ExampleApp: Application() {
  // This works fine
  @Inject lateinit var appInfo: AppInfo

  lateinit var rootGraph: AppGraph

  override fun onCreate() {
    super.onCreate()

    rootGraph = createGraphFactory<AppGraph.Factory>().create(ApplicationModule(this))
    rootGraph.inject(this)
  }
}

class ExtensionActivity: ComponentActivity() {
  // This doesn't work
  @Inject lateinit var appInfo: AppInfo

  private lateinit var extensionGraph: ExtensionGraph

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

[...]

PR #984, opened 2025-08-22
Add failing test for issue #983

erawhctim:


ZacSweers:
I'm seeing this test fail prior to reaching Metro

```
java.lang.IllegalStateException: INAPPLICABLE_LATEINIT_MODIFIER: 'lateinit' modifier is not allowed on properties of primitive types
```

erawhctim:
oops - might be fixed now? test failure looks closer to what I was seeing on the repro project 
(but the error msg isn't identical, so I still may be missing some understanding of the test suite here - bear with me)
```
Multiple Failures (2 failures)
	org.opentest4j.AssertionFailedError: Actual data differs from file content: DaggerModuleInteropWorksInGraphExtension.kt
	java.lang.IllegalStateException: //private/var/folders/ml/hvv8xpcd7zzg8py2xflx90480000gq/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Interop$DaggertestDaggerModuleInteropWorksInGraphExtension/kotlin-sources/main/DaggerModuleInteropWorksInGraphExtension.kt:(434,442): error: [Metro/MissingBinding] Cannot find an @Inject constructor or @Provides-annotated function/property for: ExtensionType

    ExtensionType is injected at
        [AppGraph.$$MetroGraph.AppGraphExtensionImpl] AppGraphExtension#inject()
    dev.zacsweers.metro.MembersInjector<MembersInjectedClass> is requested at
        [AppGraph.$$MetroGraph.AppGraphExtensionImpl] AppGraphExtension#inject()
```

erawhctim:
btw - I only see the more detailed test failure output when running the specific test case on `BoxTestGenerated` (but not when running `gw :compiler-tests:check`); does that seem right? 

I had swapped the java version back to 21 locally to match my jdk; not sure if that's at play here

ZacSweers:
Continued in #991

PR #994, opened 2025-08-25
Clean up @Extends annotation from doc

kevinguitar:


hossain-khan:
I wonder if this is one of the drawback of having versioned site! (I am guessing this update was meant to go for latest `0.6.3` release).

Now, this fix/update is only available in the snapshot build! I will think of something for these scenarios! 🤔

Issue #1012, opened 2025-09-01
Investigate: docs-site github actions workflow failed to complete `0.6.4` tagged release


hossain-khan:
The workflow run for tag event needed to be completed to publish versioned site for https://github.com/ZacSweers/metro/releases/tag/0.6.4

However, https://github.com/ZacSweers/metro/actions/runs/17368899350 failed with 

> Canceling since a higher priority waiting request for pages exists

Despite having following config to queue all to ensure nothing is missed

```yml
# Only allow one docs deployment at a time, and queue subsequent ones
concurrency:
  group: "pages"
  cancel-in-progress: false
```

hossain-khan:
Following solution could be considered, taking a look.

### Recommended Configuration

```yaml
# Only allow one docs deployment at a time per type, prioritize tagged releases
concurrency:
  group: ${{ startsWith(github.ref, 'refs/tags/') && 'pages-release' || 'pages-dev' }}
  cancel-in-progress: ${{ !startsWith(github.ref, 'refs/tags/') }}
```

This ensures:
1. ✅ **Tag deployments always succeed** (separate group, no cancellation)
2. ✅ **Regular commits can be canceled** by newer commits (efficient queue management)
3. ✅ **Tags never interfere** with or get canceled by regular deployments
4. ✅ **Multiple rapid commits** won't pile up in the queue

With this setup, your tagged releases will always deploy successfully, while regular development commits can be efficiently managed with cancellation when needed.

hossain-khan:
Okay! this workflow worked. 🙌🏽 

<img width="463" height="199" alt="Image" src="https://github.com/user-attachments/assets/e520314b-746e-4f65-838f-b8fafb8fac15" />


I think this can be used to mark old tag as latest as well, or re-publish old tag temporarily until that is deleted by cleanup script! 😂

PR #1014, opened 2025-09-01
fix: prioritize tagged releases in docs workflow concurrency

hossain-khan:
### Summary
- Use separate concurrency groups for tagged releases vs dev builds
- Tagged releases use 'pages-release' group and never get cancelled
- Dev builds use 'pages-dev' group and can be cancelled by newer builds
- Prevents tagged release docs from being cancelled by subsequent builds

Fixes issue where docs deployment was cancelled with: 'Canceling since a higher priority waiting request for pages exists'

**Related Issue:** 
* https://github.com/ZacSweers/metro/issues/1012

### DIFF
--- a/.github/workflows/docs-site.yml
+++ b/.github/workflows/docs-site.yml
@@ -23,10 +23,10 @@ permissions:
   pages: write
   id-token: write
 
-# Only allow one docs deployment at a time, and queue subsequent ones
+# Only allow one docs deployment at a time per type, prioritize tagged releases
 concurrency:
-  group: "pages"
-  cancel-in-progress: false
+  group: ${{ startsWith(github.ref, 'refs/tags/') && 'pages-release' || 'pages-dev' }}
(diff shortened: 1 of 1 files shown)


hossain-khan:
🥼 Testing this now, will open up after validation!

hossain-khan:
## ✅ Validated

Tagged release no longer gets cancelled due to `concurrency` group name collision.

### Demo

Screenshot 2025-09-01 at 10 04 26 AM

.github/workflows/docs-site.yml
```diff
@@ -23,10 +23,10 @@ permissions:
   pages: write
   id-token: write
 
-# Only allow one docs deployment at a time, and queue subsequent ones
+# Only allow one docs deployment at a time per type, prioritize tagged releases
 concurrency:
-  group: "pages"
-  cancel-in-progress: false
+  group: ${{ startsWith(github.ref, 'refs/tags/') && 'pages-release' || 'pages-dev' }}
```
ZacSweers, line 28: TIL

Issue #972, opened 2025-08-21
investigation - prepare for metro release commit does not trigger docs workflow


hossain-khan:
## Summary
I have cross referenced all the [recent commits](https://github.com/ZacSweers/metro/commits/main/) on `main` and all the recent [workflow runs](https://github.com/ZacSweers/metro/actions/workflows/docs-site.yml) and it seems like only `Prepare for release x.y.z` commits didn't trigger the workflow! 

Not exactly sure why this is happening! Will keep it open while I investigate this further! 👀 

<img width="1402" height="881" alt="Image" src="https://github.com/user-attachments/assets/1f312ba3-943b-4f9e-9aa9-e6d273c8870d" />

[Metro-commits-on-main.txt](https://github.com/user-attachments/files/21909930/Metro-commits-on-main.txt)
[Metro-docs-workflow-runs.txt](https://github.com/user-attachments/files/21909929/Metro-docs-workflow-runs.txt)

hossain-khan:
Closing in favour of https://github.com/ZacSweers/metro/pull/975 for now.

PR #1015, opened 2025-09-01
feat: add manual docs deployment workflow

hossain-khan:
## Summary
Added support for deploying docs for any specific git tag -- in the wake of on-going issues like following! 😑 
* https://github.com/ZacSweers/metro/issues/1012
* https://github.com/ZacSweers/metro/issues/972

Usage: Actions → Metro Docs Site (Manual) → Run workflow with released tag name

### DIFF
--- a/.github/workflows/docs-site-manual.yml
+++ b/.github/workflows/docs-site-manual.yml
@@ -1,0 +1,130 @@
+# Manual deployment of docs site for specific tags
+# Allows deploying documentation for any git tag on-demand
+name: Metro Docs Site (Manual)
+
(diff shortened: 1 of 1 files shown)


hossain-khan:
Testing changes, will open up when it's ready! 

Screenshot 2025-09-01 at 9 40 57 AM

hossain-khan:
Okay! this workflow worked. 🙌🏽 

Image


I think this can be used to mark old tag as latest as well, or re-publish old tag temporarily until that is deleted by cleanup script! 😂

PR #1027, opened 2025-09-05
Add Origin annotation

### LINKED ISSUE #1020: Associate generated code with the source code
### Description

Extensions for Metro using custom annotations should be treated similar to first party annotations. Imagine an annotation that generates custom bindings:
```kotlin
@MyAnnotation(AppScope::class)
class MyClass : Abc, Def

// @MyAnnotation generates this:
package any.place

interface MyClassGraph { ... }
```
It should be supported to exclude and replace code generated by custom code generators:
```kotlin
@DependencyGraph(AppScope::class, excludes = [MyClass::class])
interface TestGraph
```
The expectation here is that `MyClassGraph` is not added as super type to `TestGraph`, because `MyClass` was excluded. But that's not how it works today and the generated type `any.place.MyClassGraph` has to be excluded instead. That's not great from a user's point of view. 

Similar to the example above, `replaces` from `@ContributesTo` and `@ContributesBinding` should allow me to replace `MyClass`:
```kotlin
@ContributesTo(AppScope::class, replaces = [MyClass::class])
interface Empty

interface TestGraph
```
In this example `MyClassGraph` should not be a super type of `TestGraph`.

### kotlin-inject-anvil

kotlin-inject-anvil provided this feature out of the box with the [`@Origin`](https://github.com/amzn/kotlin-inject-anvil/blob/main/runtime/src/commonMain/kotlin/software/amazon/lastmile/kotlin/inject/anvil/internal/Origin.kt) annotation. Generated code could mark what source element generated that piece of code, e.g. our generated interface would look like this:
```kotlin
// @MyAnnotation generates this:
package any.place

@Origin(MyClass::class)
interface MyClassGraph { ... }
```
When kotlin-inject-anvil merges contributed types, it attempts to add `MyClassGraph` as super type to the component. However, through `@Origin` annotation it checks whether `MyClass` was excluded / replaced and if yes, then it doesn't add the super type and skips that file. 

This solution is simple and powerful. It doesn't require any configuration and can support any code generated by third party libraries. For users it's intuitive to use. 

I'm proposing for Metro to use a similar solution. 

### Alternatives/workarounds

_No response_

ZacSweers:
Resolves #1020


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -268,19 +272,51 @@ internal class ContributedInterfaceSupertypeGenerator(session: FirSession) :
...
+    for ((parentClassId, _) in contributions) {
+      val parentSymbol = parentClassId.toSymbol(session)?.expectAsOrNull<FirRegularClassSymbol>()
+      if (parentSymbol != null) {
+        val localTypeResolver = typeResolverFor(parentSymbol) ?: continue
+
+        parentSymbol.originClassId(session, localTypeResolver)?.let { originClassId ->
+          originToContributions.getOrPut(originClassId) { mutableSetOf() }.add(parentClassId)
+        }
+      }
+    }
```
vRallev, line 303: I can't quite tell, is this recursive for `Origin -> GeneratedAbc -> GeneratedDef`?

vRallev (approved):

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
ZacSweers: in this case it's just a straight 1:many mapping. If you have `Foo` and generate potentially multiple contributing `Foo_Impl` types, it'll track and exclude/replace all of them if `Foo` is a replacement target. So more like

```
Origin --> GeneratedAbc
       \
        -> GeneratedDef
```

vRallev: KSP still supports rounds and my thinking was that `GeneratedAbc` contains code that triggers the generation of `GeneratedDef`. `GeneratedDef` doesn't know about `Origin`, because for it the origin is `GeneratedAbc`.

ZacSweers: ahh interesting. I think I'm gonna save crossing that bridge for when someone needs it as that could be expensive in FIR :)

PR #1033, opened 2025-09-10
Fix link to Compatibility docs

ChristianKatzmann:


ChristianKatzmann:
Yes, both `https://zacsweers.github.io/metro/compatibility.` and `https://zacsweers.github.io/metro/compatibility` redirects to `https://zacsweers.github.io/metro/latest/`.

ZacSweers:
Ah gotcha, thanks. CC @hossain-khan

hossain-khan:
> Ah gotcha, thanks. CC @hossain-khan

Yes, the PR fix checks out right! because of earlier discovery the issue and opting for simpler redirect fix 👇🏽 
* https://github.com/ZacSweers/metro/pull/964

I haven't got to phase 2 of this, as it will be bit more sophisticated. 

For now, manual fix like this with `/latest/` might be much easier route.

PR #1040, opened 2025-09-11
Fix a couple of edge cases in FirAccessorOverrideStatusTransformer

kevinguitar:
This PR fixes a couple of edge cases in `FirAccessorOverrideStatusTransformer`.

### Fun shouldn't be override-compatible with val

The transformer shouldn't add override to `fun int(): Int` in this case
```kotlin
interface AppGraph {
  fun int(): Int
}

interface IntProvider {
  val int: Int
}
```

### Look up the super types of the contributed interface

The transformer should add override to `val string: String` in this case
```kotlin
interface AppGraph {
  val string: String
}

interface ValueProvider : StringProvider

interface StringProvider {
  val string: String
}
```


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/FirAccessorOverrideStatusTransformer.kt
```diff
@@ -78,20 +80,14 @@ internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
     declaration: FirDeclaration,
   ): FirDeclarationStatus {
     if (declaration !is FirCallableDeclaration) return status
-    val returnType = declaration.returnTypeRef.coneTypeOrNull ?: return status
+    declaration.returnTypeRef.coneTypeOrNull ?: return status
+    declaration.symbol.callableId?.callableName ?: return status
```
kevinguitar, line 87: Since we don't need the fields anymore, I lifted the null check here, but I'm not sure if they're still needed

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/FirAccessorOverrideStatusTransformer.kt
```diff
@@ -78,20 +80,14 @@ internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
...
-
     var needsOverride = false
-    for (superType in containingClass.getSuperTypes(session, recursive = false)) {
+    for (superType in containingClass.getSuperTypes(session)) {
       val classSymbol = superType.toClassSymbol(session) ?: continue
-      if (
-        !classSymbol.name.asString().startsWith(Symbols.StringNames.METRO_CONTRIBUTION_NAME_PREFIX)
-      ) {
-        continue
-      }
```
kevinguitar, line 94: This is the change I'm concerned about.

As I see, this is the only way to make override-compatible check right, but it obviously is heavier now because we look up all super types. Let me know if you see any issues with this fix.

ZacSweers: this seems fair. One thing we could do is cache checked callable IDs while searching (i.e. if it's declared in a subtype, no need to compare against it in the parent type)?

kevinguitar: Yeah it sounds good as an optimization, I remember I saw something similar in the codebase somewhere. I'll look into it during weekend 👍

kevinguitar: Not sure if I got you right, I've made an attempt here https://github.com/ZacSweers/metro/pull/1040/commits/66a9f392e403fb32608ed07739c915dcf8003041

ZacSweers: Hmm I'm not sure it's safe to cache at the class level as I'm not sure what the lifecycle of this is. I was thinking just caching locally when recursing up supertypes

kevinguitar: Ohh fair point, let me revisit

kevinguitar: Let me wrap up my mind, a bit lost about what to cache in the function level

> if it's declared in a subtype, no need to compare against it in the parent type

For example
```kotlin
interface Child : Parent {
  override val int: Int
}

interface Parent {
  val int: Int
}
```
Do you mean that `Child.int` will be checked first, and we skip checking `Parent.int` in this case?

ZacSweers: right 👍, as in that case we've already confirmed the matching declaration and can skip checking it in any supertypes

kevinguitar: Gotcha 👍

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/FirAccessorOverrideStatusTransformer.kt
```diff
@@ -119,4 +122,18 @@ internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
     if (!needsOverride) return status
     return status.copy(isOverride = true)
   }
+
+  private data class Accessor(
+    val name: Name,
+    val isFunction: Boolean,
+    val returnType: ConeKotlinType
+  )
```
kevinguitar, line 131: Introduced a local class to represent the accessor information, which is used for storing in a set and comparison.

kevinguitar:
Btw, noticed something while debugging the transformer. Looks like it runs through the FIR elements that metro generates as well, do you think we can safely exclude them? If so, I assume there is a way to check whether a `FirDeclaration` is generated by metro or not?

```
Transforming status for FirNamedFunctionSymbol /AppGraph.int
Transforming status for FirRegularPropertySymbol /AppGraph.string
Transforming status for FirPropertyAccessorSymbol special/accessor    <---
Transforming status for FirBackingFieldSymbol <local>/field           <---
Transforming status for FirRegularPropertySymbol /AppGraph.boolean
Transforming status for FirPropertyAccessorSymbol special/accessor    <---
Transforming status for FirBackingFieldSymbol <local>/field           <---
```

ZacSweers:
> Btw, noticed something while debugging the transformer. Looks like it runs through the FIR elements that metro generates as well, do you think we can safely exclude them? If so, I assume there is a way to check whether a `FirDeclaration` is generated by metro or not?
> 
> ```
> Transforming status for FirNamedFunctionSymbol /AppGraph.int
> Transforming status for FirRegularPropertySymbol /AppGraph.string
> Transforming status for FirPropertyAccessorSymbol special/accessor    <---
> Transforming status for FirBackingFieldSymbol <local>/field           <---
> Transforming status for FirRegularPropertySymbol /AppGraph.boolean
> Transforming status for FirPropertyAccessorSymbol special/accessor    <---
> Transforming status for FirBackingFieldSymbol <local>/field           <---
> ```

Yeah we can ignore those since we know we're not generating new declarations there that we're interested in 👍. You can check the `key` of the declaration symbol (there are a bunch of predefined ones in `Keys.kt`)

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/FirAccessorOverrideStatusTransformer.kt
```diff
@@ -119,4 +122,18 @@ internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
...
+
+  @OptIn(SymbolInternals::class)
+  private val FirCallableSymbol<*>.accessor: Accessor
+    get() = Accessor(
```
ZacSweers: Can we avoid the symbol internals call? Should be able to check if the symbol is an `FirNamedFunctionSymbol` or something similar

kevinguitar: Let me try

kevinguitar: You're right! Done here https://github.com/ZacSweers/metro/pull/1040/commits/98da5924e71bf404dc238d8d453953930387560d

ZacSweers:
It makes sense that those symbols have Source keys as they are defined in source, I maybe misunderstood what you meant earlier as I thought you meant declarations generated by metro in some earlier round of FIR gen

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/FirAccessorOverrideStatusTransformer.kt
```diff
@@ -38,6 +42,7 @@ internal class FirAccessorOverrideStatusTransformer(session: FirSession) :
     if (declaration !is FirCallableDeclaration) return false
     if (declaration is FirConstructor) return false
     if (declaration is FirField) return false
+    if (declaration is FirBackingField) return false
```
kevinguitar, line 45: > You can check the key of the declaration symbol

Unfortunately, the suggested approach doesn't work. Both `special/accessor` and `<local>/field` have `Source` origin instead of `Plugin`.

I excluded `FirBackingField` here as a whole, it should be safe because backing field isn't allowed in interfaces.

But `special/accessor` is trickier, I wrote some message here in the commit message here. https://github.com/ZacSweers/metro/pull/1040/commits/0d8a20dacd48564e54a72e09ac975163dbbfa392

ZacSweers (approved):
Lgtm. I'll ask around if there's a more optimal API for checking these

kevinguitar:
Thanks Zac! 🙇‍♂️

Issue #42, opened 2024-12-14
Consider implementing SwitchingProviders


ZacSweers:
Dagger does this to avoid classloading as an optimization, essentially what the fastinit option does

ZacSweers:
Startup perf results of a large (1000 module) benchmark

# Disabled

## Startup Benchmark Results: z/fastInit

**Date:** Mon Jan 19 12:24:11 EST 2026
**Module Count:** 1000
**Modes:** metro
**Commit:** 70a01336 Cleanup

### JVM Benchmarks (JMH)

Graph creation and initialization time (lower is better):

| Framework | Time (ms) | Alloc (KB/op) | vs Metro |
|-----------|-----------|---------------|----------|
| metro | 0.369 | 177.54 | baseline |

### JVM Benchmarks - R8 Minified (JMH)

Graph creation and initialization time with R8 optimization (lower is better):

| Framework | Time (ms) | Alloc (KB/op) | vs Metro R8 |
|-----------|-----------|---------------|-------------|
| metro | 0.358 | N/A | baseline |

### Android Benchmarks (Microbenchmark)

Graph creation and initialization time on Android (lower is better):

| Framework | Time (ms) | vs Metro |
|-----------|-----------|----------|
| metro | 1.364 | baseline |


### Binary Metrics

#### Pre-Minification Component Classes

| Framework | Fields | Methods | Shards | Size (KB) |
|-----------|--------|---------|--------|-----------|
| metro | 1916 | 2080 | 0 | 9620.5 |

#### R8-Minified JAR

| Framework | JAR Size (KB) | Classes | Methods | Fields |
|-----------|---------------|---------|---------|--------|
| metro | 5078.1 | 7396 | 11162 | 1858 |

#### Android APK

| Framework | APK Size (KB) | DEX Size (KB) | DEX Classes | DEX Methods | DEX Fields |
|-----------|---------------|---------------|-------------|-------------|------------|
| metro | 785.9 | 554.2 | 779 | 4201 | 1936 |

# Enabled

## Startup Benchmark Results: z/fastInit

**Date:** Mon Jan 19 12:31:34 EST 2026
**Module Count:** 1000
**Modes:** metro
**Commit:** 419e9703 More debug

### JVM Benchmarks (JMH)

Graph creation and initialization time (lower is better):

| Framework | Time (ms) | Alloc (KB/op) | vs Metro |
|-----------|-----------|---------------|----------|
| metro | 0.351 | 221.30 | baseline |

[...]

PR #1043, opened 2025-09-12
WIP: Graph sharding

izantech:
**Summary**

This PR will add support for graph sharding and switching providers in Metro's compiler plugin. Large dependency graphs on the JVM can cause generated classes and methods to exceed the bytecode limits enforced by the JVM (e.g. 64KB per method), resulting in `ClassTooLargeException`/`MethodTooLargeException` during compilation.  Dagger solves this problem by splitting component implementations into multiple "shards" and delegating between them.  This change lays the groundwork for a similar solution in Metro.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/graph/GraphSharding.kt
```diff
@@ -0,0 +1,144 @@
...
+    val inDegree = IntArray(componentCount)
+    dag.values.flatten().forEach { comp -> inDegree[comp]++ }
+
+    // Use a priority queue to ensure deterministic ordering when multiple components are ready.
+    val ready: PriorityQueue<Int> = PriorityQueue()
+    for (i in 0 until componentCount) {
+      if (inDegree[i] == 0) ready.add(i)
+    }
+
+    // Kahn's algorithm for topologically sorting the component DAG.  The resulting list contains
```
ZacSweers, line 78: Can we reuse anything from `topologicalSort.kt` rather than reimplement here?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/graph/GraphSharding.kt
```diff
@@ -0,0 +1,144 @@
...
+   * @param threshold the maximum number of bindings allowed per shard
+   */
+  internal fun <K : Any, B : Any> computeShardsFromBindings(
+    bindings: Map<K, B>,
+    dependenciesOf: (B) -> Iterable<K>,
+    threshold: Int = DEFAULT_BINDINGS_PER_SHARD
+  ): List<List<K>> {
+    // Build a sorted adjacency map of keys to their dependencies.  We use a TreeSet/TreeMap so
+    // Tarjan's SCC algorithm runs deterministically.
+    val adjacency: SortedMap<K, java.util.SortedSet<K>> = java.util.TreeMap<K, java.util.SortedSet<K>>()
```
ZacSweers, line 132: Were these fully-qualified imports generated by an AI tool?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ShardingExtensions.kt
```diff
@@ -0,0 +1,32 @@
...
+ * implements a similar approach via [GraphSharding].  Consumers can query this extension to
+ * obtain a list of shards representing contiguous subsets of the binding graph.  Each inner
+ * list contains the [IrTypeKey]s that belong to that shard in dependency order.
+ *
+ * The shards are computed on demand using the [GraphSharding.computeShardsFromBindings] helper.
+ * For deterministic output, the adjacency is constructed from the bindings snapshot of the
+ * [IrBindingGraph] and sorted.  This property does not cache its result, so repeated calls
+ * recompute the sharding; callers should memoize if necessary.
+ */
+internal val IrBindingGraph.shards: List<List<IrTypeKey>>
```
ZacSweers, line 23: nit: given that this seems to be a fairly expensive operation, I'd rather this be a function like `computeShards()`

izantech:
> This is exciting! A few early questions
> 
> * Given the amount of work that needs to go into this, could we target this into a non-main branch and iterate there? I'm happy to make a `shards` branch for that case if you're ok with it
> * While the JVM does have those limitations, do we have some classloading benefits here to shard on non-jvm anyway? Or at the very least, is there any real harm to doing this for the sake of consistency of generated code?
> * Do you know how this would impact #42?
> * Can we add tests for `GraphSharding`?
> * Was some of this generated by AI? It's ok if so, I just would like to know so I can tune my review accordingly. Asking as I see some of the fully-qualified types inline, which is a common artifact of AI code gen in my experience

Hi there! Sorry about the late response.
- Absolutely! I don't know how long is it going to take me to finish it, so that would be a good idea.
- I haven't thought about it, but I can look into it.
- Actually since I'm trying to replicate Dagger that would be also part of this PR.
- For sure! I intend to add tests for all new features related to graph sharding and switching providers.
- Yes, I'm currently using GPT5 and Claude Code to help me, because I'm really new to K2 compiler plugins.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/Symbols.kt
```diff
@@ -501,6 +504,10 @@ internal class Symbols(
     val doubleCheckCompanionObject by lazy { doubleCheck.owner.companionObject()!!.symbol }
     val doubleCheckProvider by lazy { doubleCheckCompanionObject.requireSimpleFunction("provider") }
 
+    // SingleCheck support for unscoped bindings (lighter weight than DoubleCheck)
+    // Returns null if SingleCheck is not available in the runtime
+    abstract fun singleCheckProviderOrNull(): IrSimpleFunctionSymbol?
```
ZacSweers, line 509: Metro explicitly doesn't support single-check, was this necessary? I'd like to minimize scope creep since this already going to be a large set of changes

compiler-tests/src/test/data/box/sharding_init_chunking_guarantee.kt
```diff
@@ -0,0 +1,154 @@
+// COMPILER_OPTIONS: -Xplugin-option=metro:sharding.enabled=true -Xplugin-option=metro:sharding.keysPerShard=30
```
ZacSweers, line 1: There is a Directives API you should use instead of raw compiler options

compiler-tests/src/test/data/box/sharding_init_chunking_guarantee.kt
```diff
@@ -0,0 +1,154 @@
...
+// TEST_TARGET: jvm
```
ZacSweers, line 2: All targets in compiler-tests currently are JVM only, I'm not sure this flag is doing anything?

compiler-tests/src/test/data/box/sharding_sanity.kt
```diff
@@ -0,0 +1,147 @@
...
+  assertTrue(
+    boundInstanceFields.isNotEmpty(),
+    "Graph should have field for bound instance"
+  )
+
+  return "OK"
+}
+
+// Helper assertions
+fun assertNotNull(value: Any?, message: String = "Value should not be null") {
```
ZacSweers, line 131: All these assertions from kotlin.test.* are implicitly available as the `kotlin.test.*` import is added automatically

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -175,6 +182,48 @@ import org.jetbrains.kotlin.name.SpecialNames
 internal class DependencyGraphFirGenerator(session: FirSession) :
   FirDeclarationGenerationExtension(session) {
 
+  @OptIn(DirectDeclarationsAccess::class)
+  private fun createSwitchingProviderSkeleton(
+    session: FirSession,
+    owner: FirClassSymbol<*>,
+    name: Name
+  ): FirClassLikeSymbol<*> {
+    // Check if already exists (idempotent on incremental builds)
```
ZacSweers, line 191: could you unpack this comment for me more? Were you seeing this type already existing before?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -175,6 +182,48 @@ import org.jetbrains.kotlin.name.SpecialNames
 internal class DependencyGraphFirGenerator(session: FirSession) :
   FirDeclarationGenerationExtension(session) {
 
+  @OptIn(DirectDeclarationsAccess::class)
```
ZacSweers, line 185: Can you avoid this? There are newer APIs for this

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
ZacSweers, line 1: We only need to generate FIR for declarations that need to be visible (in metadata) to other compilations. That doesn't seem to be the case here, can we restrict this all to just IR?

compiler-tests/src/test/data/box/sharding_init_chunking_guarantee.kt
izantech: Got it, thanks!

compiler-tests/src/test/data/box/sharding_sanity.kt
izantech: I missed those, thanks!

ZacSweers:
@IzanRodrigo there is a lot of unnecessary AI-generated stuff in this PR, for example the project status markdown doc. While I am accepting of code generated with AI, I'd like to ask that you do more review before pushing it up here to clean up the extra cruft and avoid the duplicated code.

It would also be helpful if you would document the design of this feature into a markdown doc somewhere, both for review purposes as well as future readers to understand its implementation.

izantech:
> @IzanRodrigo there is a lot of unnecessary AI-generated stuff in this PR, for example the project status markdown doc. While I am accepting of code generated with AI, I'd like to ask that you do more review before pushing it up here to clean up the extra cruft and avoid the duplicated code.
> 
> It would also be helpful if you would document the design of this feature into a markdown doc somewhere, both for review purposes as well as future readers to understand its implementation.

Understood! I'll address the comments in the PR and keep working on my fork until a more detailed review makes sense.
Thanks for the comments 🙂

ZacSweers:
Sounds good, thanks!

izantech:
Working on addressing the comments and finishing the implementation here: https://github.com/IzanRodrigo/metro/pull/3
I'll create a new PR once is working, sorry about the mess here!

Issue #543, opened 2025-06-05
Investigate supporting re-providing of assisted types via their factories


ZacSweers:
A test for this

```kotlin
// Edge case of explicitly providing a constructor-injected assisted type
// overriding the one on the graph *via its factory* and thus bypassing
// the error reporting we do for doing this normally
@Inject
class Example(@Assisted val input: String) {
  @AssistedFactory
  fun interface Factory {
    fun create(input: String): Example
  }
}

@DependencyGraph
interface AppGraph {
  // Request the assisted type
  val example: Example

  @Provides
  val string: String get() = "Hello"

  @Provides
  fun provideExample(factory: Example.Factory, input: String): Example = factory.create(input)
}

fun box(): String {
  val graph = createGraph<AppGraph>()
  assertNotNull(graph.example)
  return "OK"
}
```

Currently this fails with a cycle error (correctly), because the factory binding depends on the assisted class. This is because the assisted class participates in the graph like any other binding.

A solution for this would be to do something similar to what metro does with multibindings - add a synthetic generated (but deterministic) qualifier to the assisted type on the graph that is unlikely for the user to request. So we're treat it more like

```kotlin
@Named("assisted type for test.Example") @Inject class Example
```

Then if you re-provide an `Example` instance later it won't collide with the original. Metro would report an error like this if you tried to depend on it separately without an explicit provision

```
e: ExampleClass.kt:19:3 [Metro/InvalidBinding] 'test.ExampleClass' uses assisted injection and cannot be injected directly. You must inject a corresponding @AssistedFactory type instead.

(Hint)
It looks like the @AssistedFactory for 'test.ExampleClass' is 'test.ExampleClassFactory'.
```

It'd be a little awkward because of course these types could have their own qualifiers too, and we'd need to tie link those back together. You could also have assistedfactory-created types that don't have any assisted parameters 😅

ZacSweers:
If we do this, we can also skip the new reverse adjacency validation introduced in #542 and defer that check to just a "similar binding" lookup at error-time

ZacSweers:
Looks like dagger just prohibits this. Seems right

```kotlin
@Component
interface UnitGraph {
  val foo: Foo

  @Provides
  fun provideFoo(factory: Foo.Factory): Foo {
    return factory.create("UnitGraph")
  }
}

class Foo @AssistedInject constructor(
  @Assisted str: String,
) {
  @AssistedFactory
  interface Factory {
    fun create(str: String): Foo
  }
}
[...]

PR #1054, opened 2025-09-16
A couple of failing tests for AssistedFactory usages

kevinguitar:
A few folks from my team started to see some weird runtime crashes, such as 

```
Caused by: java.lang.ClassCastException: dev.zacsweers.metro.internal.DoubleCheck cannot be cast to com.bandlab.share.dialog.template.api.ClipMakerAssetRepository$$$MetroFactory
  at com.bandlab.bandlab.AppGraph$$$MetroGraph$TrendingFeedPageGraphExtensionImpl.init2(Unknown Source:1349)
  at com.bandlab.bandlab.AppGraph$$$MetroGraph$TrendingFeedPageGraphExtensionImpl.<init>(Unknown Source:52)
  at com.bandlab.bandlab.AppGraph$$$MetroGraph.create(AppGraph.kt:11)
```

I found out the problem with this crash is because of missing dependencies in the graph that uses the assisted factory, so I went ahead to create a test case.

But there is a bizarre thing, if you look at the `SimpleAssistedFactory.kt` test case, it's a super simple case, but it still results in a runtime crash 😰  

For the problem my devs are facing, it's reproducible in `AssistedFactoryDependenciesAreValidated.kt`.

### DIFF
--- a/compiler-tests/src/test/data/box/inject/assisted/AssistedFactoryDependenciesAreValidated.kt
+++ b/compiler-tests/src/test/data/box/inject/assisted/AssistedFactoryDependenciesAreValidated.kt
@@ -1,0 +1,35 @@
+@DependencyGraph(AppScope::class)
+interface AppGraph
+
+@DependencyGraph(Unit::class)
+interface UnitGraph {
+  val foo: Foo
+
+  @Provides
+  fun provideFoo(factory: Foo.Factory): Foo {
+    return factory.create("UnitGraph")
+  }
+}
+
+@Inject
+class Foo(
+  @Assisted str: String,
+  bar: Bar // <-- this should fail because it doesn't exist in UnitGraph
+) {
+  @AssistedFactory
+  interface Factory {
+    fun create(str: String): Foo
+  }
+}
+
+interface Bar
+
+@Inject
+@ContributesBinding(AppScope::class)
+class BarImpl : Bar
+
+fun box(): String {
+  val unitGraph = createGraph<UnitGraph>()
+  assertNotNull(unitGraph.foo)
+  return "OK"
+}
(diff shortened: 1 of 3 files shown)


ZacSweers:
When you say "started", did this start in a particular release? Would help if you could pinpoint a start time to narrow down where it was introduced if its a regression

kevinguitar:
Let me check that a bit more, so far we tried building with 0.6.4 and it's also reproducible there.

ZacSweers:
Or better question - can you pinpoint if it was a regression or always existed?

ZacSweers:
Oh, this is a cycle. This is essentially a variant of #543 right?

kevinguitar:
Hmm, it's a bit different though, the `String` isn't available in the graph, so I assume it's not a cycle? Coz the only way to create it is assisted

ZacSweers:
Foo requires a Foo.Factory to be instantiated, but technically Foo.Factory requires a Foo binding in the graph and it's just the only thing that can construct this instance. I'm fairly certain it's under that same umbrella anyway as they're treated a little funny at the moment

ZacSweers:
Yeah dagger prohibits your example, I think we'll do the same https://github.com/ZacSweers/metro/issues/543#issuecomment-3297169751

kevinguitar:
Ah I see it now, let me update this PR to reflect the problem my devs ran into

kevinguitar:
Decided to close this one for now to reduce noise, will create a new one if I can reproduce the exact stacktrace our devs are seeing 🙏  Thank you Zac for quick responses

kevinguitar:
I can confirm the case our devs are facing is the same situation - the cycle mentioned above

PR #1075, opened 2025-09-20
Ensure graph extension classes implement contributed interfaces

### LINKED ISSUE #1074: Graph extension classes do not implement contributed interfaces
### Description

Currently, contributed interfaces get added as supertypes for the generated Impl class for a graph extension, but not the graph extension class itself. This means that you can only perform comparisons for a contributed interface to a live instance of a graph extension but not the graph extension class.

Something we do regularly is use reflection to find the graph extension factory method we want based on return type and whether it maps to a contributed type. E.g. imagine you have multiple apps that each define their own graph and graph extensions for the same scopes, like graph A and extension B. Then you have various feature modules that contribute interfaces to B's scope. The feature modules do not know about B directly, but can look up factory methods where the contributed interface is assignable from the method return type (B) in order to call and create the extension.

### Self-contained Reproducer

[compiler-tests/src/test/data/box/contributesgraphextension/WithContributedAccessorInterface.kt](https://github.com/ZacSweers/metro/pull/1075/files#diff-ff68647320652ad6721dda5f0f369ab64fe7c1e6bc4446a1e9cfdcbbbc26dc07)

### Metro version

0.7.0-SNAPSHOT

### Context

_No response_

JoelWilcox:
- Previously, only instances of an extension would implement the contributed interfaces.
- This is mostly important for use-cases where reflection is used to compare graph methods's return types against contributed types.
- As a more concrete example: imagine you have multiple apps that each define their own graph and graph extensions for the same scopes, like graph A and extension B. Then you have various feature modules that contribute interfaces to B's scope. The feature modules do not know about B directly, but can look up factory methods where the contributed interface is assignable from the method return type (B) in order to call and create the extension.
- Fixes #1074

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changelog
 - **Enhancement**: Simplify assisted factory impl class generation by moving it entirely to IR.
 - **Fix**: Register `MetroDiagnostics` in FIR.
 - **Fix**: When transforming FIR override statuses, check all supertypes and not just immediate supertype.
+  **Fix:** Ensure graph extension classes implement contributed interfaces. Previously, only instances of an extension would implement the contributed interfaces.
 
 0.6.6
 -----
--- a/compiler-tests/src/test/data/box/contributesgraphextension/WithContributedAccessorInterface.kt
+++ b/compiler-tests/src/test/data/box/contributesgraphextension/WithContributedAccessorInterface.kt
@@ -1,0 +1,49 @@
+// https://github.com/ZacSweers/metro/issues/1074
+// Use-case: ensure that the graph extension class has its supertypes updated with contributed
+// accessor interfaces, not just the implementation used under the hood.
+// MODULE: lib
+package test
+
+abstract class ChildScope
+
+@GraphExtension(ChildScope::class)
+interface ChildGraph {
+
+  @GraphExtension.Factory
+  @ContributesTo(AppScope::class)
+  interface Factory {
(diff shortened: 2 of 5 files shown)


ZacSweers:
I don't think this is gonna work, because contributed supertypes are merged only at the creation of the extension impl and not the extension declaration. Those could be different depending on where they are merged. I don't see how this would have been viable with ContributesSubcomponent before either. Am I missing something?

The ways this can break down
- not all contributed interfaces are visible to the declaration
- a declaration elsewhere may exclude a contribution that is visible to the declaration, but by then it's too late to remove it

JoelWilcox:
These are great catches! Did some testing to verify and confirmed that Anvil is affected by both of the issues you listed. I would consider the first point more of a reasonable limitation that could be documented (basically make sure a contributed interface is on the graph extension's compile classpath, not just the merging module, if you care about using this type of lookup functionality). But the second point for replacements and exclusions is pretty bad.

We haven't really been affected by these issues since we're always looking up the factory methods by interfaces that are visible to the graph extension declarations, and we never exclude these types of interfaces. I'm thinking about what alternatives we could do here on the consuming side if this isn't an option. In the meantime, I'm curious would you be open to supporting this when Dagger/Anvil interop is enabled to match the existing functionality?

JoelWilcox:
After thinking on this a bit more, considering that this functionality didn't fully work even in Anvil and I have a pretty simple alternative approach that looks like it should work, I'll go ahead and close this.

Issue #1079, opened 2025-09-23
Align return types for Metro with Dagger


vRallev:
### Description

Imagine having a contributed subcomponent with Dagger and Anvil like this:
```kotlin
@ContributesSubcomponent(
  scope = ActivityScope::class,
  parentScope = AppScope::class
)
interface ActivitySubcomponent {

  @ContributesTo(AppScope::class)
  interface ParentComponent {
    fun createActivitySubcomponent(): ActivitySubcomponent
  }
}
```
Then the final app component generated by the Dagger annotation processor will implement the `ParentComponent` interface roughly like this (pseudo code):
```kotlin
class DaggerAppComponent {
  override fun createActivityComponent(): DaggerActivitySubcomponent = ...
}
```
The important piece is that Dagger changes the return type for the overridden function to the concrete generated type `DaggerActivitySubcomponent` and doesn't keep using the return type `ActivitySubcomponent` declared in the contributed interface.

Metro doesn't have this behavior and keeps the original type `ActivitySubcomponent` as return type for the overridden function. For interop reasons it would be great to preserve the behavior from Dagger.

vRallev:
In practice, this shouldn't make a big difference, unless you rely on reflection and inspect the type returned by the factory method 🙃

ZacSweers:
Sounds good to me, wanna send a PR?

vRallev:
Yes, happy to in the coming days. I wanted to hear your opinion first. Please wait for my PR :-)

ZacSweers:
Done with #1094 behind a flag

PR #1094, opened 2025-09-25
Align return types for Metro graph factories with Dagger

vRallev:
Change the return type of generated factory functions for graphs and graph extensions from the interface type to the concrete generated class types. This behavior is aligned with Dagger.

See #1079

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExpressionGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExpressionGenerator.kt
@@ -279,6 +279,12 @@ private constructor(
               parentTracer,
             )
 
+          if (options.enableGraphImplClassAsReturnType) {
+            // This is probably not the right spot to change the return type, but the IrClass
+            // implementation is not exposed otherwise.
+            binding.accessor.returnType = extensionImpl.defaultType
+          }
+
           val ctor = extensionImpl.primaryConstructor!!
           val instanceExpression =
             irCallConstructor(ctor.symbol, node.sourceGraph.typeParameters.map { it.defaultType })
--- a/compiler-tests/src/test/data/box/contributesgraphextension/ChildGraphIsUsedAsReturnType.kt
+++ b/compiler-tests/src/test/data/box/contributesgraphextension/ChildGraphIsUsedAsReturnType.kt
@@ -1,0 +1,44 @@
+// MODULE: lib
+@GraphExtension(String::class)
+interface ChildGraph {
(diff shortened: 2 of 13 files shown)


compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/GraphExtensionTest.kt
```diff
@@ -601,4 +603,46 @@ class GraphExtensionTest : MetroCompilerTest() {
       assertThat(childGraph.callProperty<Int>("int")).isEqualTo(1)
     }
   }
+
+  @Test
+  fun `generated child graph factory functions return the generated child graph type`() {
```
vRallev: @ZacSweers You mentioned that this spot for this test is likely wrong. I can create an explicit IR dump test if you like, but I already had to update some. I wanted to keep this test nonetheless to document the behavior around the two functions available via reflection.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExpressionGenerator.kt
```diff
@@ -279,6 +279,10 @@ private constructor(
               parentTracer,
             )
 
+          // This is probably not the right spot to change the return type, but the IrClass
+          // implementation is not exposed otherwise.
+          binding.accessor.returnType = extensionImpl.defaultType
```
vRallev: @ZacSweers You mentioned to hide this behavior behind a flag. Could you point me to any other flag I can look at as an example?

ZacSweers: take a look at `MetroOptions` and how they are wired up

compiler/src/test/kotlin/dev/zacsweers/metro/compiler/transformers/GraphExtensionTest.kt
ZacSweers: you can still do this in the box function and just do the reflection there. The goal is to eventually remove all the kotlin-compile-testing tests

vRallev: Done 👍🏻

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExpressionGenerator.kt
vRallev: Done 👍🏻

ZacSweers (approved):

PR #1102, opened 2025-09-27
fixed (potentially): `gh-pages` re-synched to workaround concurrent commit

hossain-khan:
It seems like on rare-occassion, the right after release the `main` branch commit can make commit on `gh-pages` before the release tag workflow finishes.

For example the `0.6.8` release task:
* https://github.com/ZacSweers/metro/actions/runs/18047435530/job/51361485231

This should ideally fix the issue with outdated branch before `mike` deployment

### DIFF
--- a/.github/workflows/docs-site.yml
+++ b/.github/workflows/docs-site.yml
@@ -136,6 +136,12 @@ jobs:
           rm -rf /tmp/docs-site-validation
           echo "Docs validation completed successfully"
 
+      - name: Re-sync gh-pages branch before deployment
+        run: |
+          # Safely update local gh-pages to match remote state to avoid push conflicts
+          git fetch origin gh-pages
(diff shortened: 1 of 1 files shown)


hossain-khan:
Considered few approaches, this one seems simple and safer!

hossain-khan:
By the way, the underlying reason why this happens because [earlier](https://github.com/ZacSweers/metro/pull/1014) I have isolated tagging event into separate group
* **`pages-release`**
* **`pages-dev`**

As seen below. However, it was done to fix tagged release issue. So, I don't think we can go back! So, this workaround might be okay.

```yml
# Only allow one docs deployment at a time per type, prioritize tagged releases
concurrency:
  group: ${{ startsWith(github.ref, 'refs/tags/') && 'pages-release' || 'pages-dev' }}
  cancel-in-progress: ${{ !startsWith(github.ref, 'refs/tags/') }}
```

PR #1121, opened 2025-10-06
Fix mixed minimized deferred cycle logic + some other tuneups

kevinguitar:
Recently, one of our devs ran into a StackOverflowError at runtime in a valid cycle. The use case we have is to inject the real implementation in a fake and replace the real one with the fake. This way, the fake can toggle the real behavior if needed.

~I couldn't reproduce the error in test so far, but I noticed the replacer doesn't seem to be validated correctly.~ I came up with a test case with the following error:
```
Caused by: java.lang.IllegalStateException: Hard cycle remained inside SCC after removing selected soft edges
	at dev.zacsweers.metro.compiler.graph.TopologicalSortKt.sortVerticesInSCC(topologicalSort.kt:460)
	at dev.zacsweers.metro.compiler.graph.TopologicalSortKt.topologicalSort$lambda$9(topologicalSort.kt:243)
	at dev.zacsweers.metro.compiler.tracing.TracerKt.trace(Tracer.kt:86)
	at dev.zacsweers.metro.compiler.tracing.TracerKt.traceNested(Tracer.kt:79)
	at dev.zacsweers.metro.compiler.tracing.TracerKt.traceNested$default(Tracer.kt:73)
	at dev.zacsweers.metro.compiler.graph.TopologicalSortKt.topologicalSort(topologicalSort.kt:234)
	at dev.zacsweers.metro.compiler.graph.MutableBindingGraph.sortAndValidate$lambda$1(BindingGraph.kt:255)
	at dev.zacsweers.metro.compiler.tracing.TracerKt.trace(Tracer.kt:86)
```

### DIFF
--- a/compiler-tests/src/test/java/dev/zacsweers/metro/compiler/BoxTestGenerated.java
+++ b/compiler-tests/src/test/java/dev/zacsweers/metro/compiler/BoxTestGenerated.java
@@ -378,12 +378,6 @@ public class BoxTestGenerated extends AbstractBoxTest {
     }
 
     @Test
-    @TestMetadata("HardCycleWithReplace.kt")
-    public void testHardCycleWithReplace() {
-      runTest("compiler-tests/src/test/data/box/cycles/HardCycleWithReplace.kt");
-    }
-
-    @Test
     @TestMetadata("LongCycle.kt")
     public void testLongCycle() {
       runTest("compiler-tests/src/test/data/box/cycles/LongCycle.kt");


ZacSweers:
> the replacer doesn't seem to be validated correctly

I'm not sure I understood what this means, can you elaborate?

ZacSweers:
Some notes while looking: The fake impl is still a hard cycle for `A -> FakeA -> B -> A`. I expect this to fail and it does if you remove the `RealA` dependency:

```
error: [Metro/DependencyCycle] Found a dependency cycle while processing 'CycleGraph'.
Cycle:
    B --> FakeA --> A --> B

Trace:
    B is injected at
        [CycleGraph] FakeA(…, b)
    FakeA is injected at
        [CycleGraph] FakeA.$$MetroContributionToUnit#bindsAsA(…, instance)
    A is injected at
        [CycleGraph] B(…, a)
    B is injected at
        [CycleGraph] FakeA(…, b)
    ...
```

The error message should be better though.

If you remove the B dep and only have the RealA dep, it correctly compiles.

The replacement nature here doesn't have any bearing, same results happen without replacing anything

kevinguitar:
ah sorry, validate is not the right word, I meant when the replacer contains real impl, dependencies seem not to be sorted correctly. Not 100% sure but I think the stackoverflow crash my dev saw is related to this.

kevinguitar:
Let me spend a bit more time today and see if I can reproduce the stackoverflow error

kevinguitar:
I managed to reproduce it! But an interesting thing is that even if you comment out the replacer, it crashes too. It is not the case for us for some reason.

**Edit**: Filed a separate PR here https://github.com/ZacSweers/metro/pull/1125

Issue #460, opened 2025-05-24
Hint generation is unsupported on non-jvm platforms


r0adkll:
### Description

I have a case where contributed bindings are not being found when compiling my iOS graph in a separate module. Take the example

```kotlin
// feature/commonMain/…/FeatureDependency.kt
interface FeatureDependency

@ContributesBinding(AppScope::class)
@Inject
class FeatureDependencyImpl : FeatureDependency 

// shared/iosMain/…/IosAppGraph.kt
@DependencyGraph(AppScope::class)
interface IosAppGraph : AppGraph {
  val featureDep: FeatureDependency
}
```

When running `gw :shared:compileKotlinIosSimulatorArm64` it will fail with 

```
e: file:///…/IosAppGraph.kt:13:3 [Metro/MissingBinding] Cannot find an @Inject constructor or @Provides-annotated function/property for: com.example.feature.FeatureDependency

    com.example.feature.FeatureDependency is requested at
        [com.example.metrobugreproduction.IosAppGraph] com.example.metrobugreproduction.IosAppGraph#featureDep

```

### Self-contained Reproducer

See https://github.com/r0adkll/MetroBugs/tree/dh/ios-compilation

Steps to reproduce:
1. Run `./gradlew :shared:compileKotlinIosSimulatorArm64` 
2. See `MissingBinding` error

### Metro version

0.3.2

### Context

I ran into this in my migration where I have iOS specific app graph stuff for exposing/creating the compose UIViewController for the swift code and when migrating hit a bunch of these missing binding exceptions.

ZacSweers:
This is expected right now unfortunately. Hints cannot currently be generated in IR in native targets due to https://youtrack.jetbrains.com/issue/KT-58886. Buuuuut that should hopefully be addressed soon 🤞. Right now they only work on native targets within the same compilation. Sorry for the confusion, I thought I had this documented somewhere but seems not yet

gabrielittner:
Am I assuming correctly that this will also cause `@ContributesGraphExtension` to not properly work?

ZacSweers:
Correct yeah. In the mean time you _should_ be able to make those still work by manually extending the contributed graph's factory class. i.e.

```kotlin
@ContributesGraphExtension(...)
interface LoggedInGraph {
  @ContributesGraphExtension(AppScope::class)
  interface Factory { ... }
}

@DependencyGraph(AppScope::class)
interface AppGraph : LoggedInGraph.Factory {
[...]

PR #1122, opened 2025-10-06
fix: #460 - FIR Analysis Diagnostics Hints

baksha97:
# Background
I've been working on my own compiler plugin and I've been very grateful to have such a comprehensive resource such as Metro. I believe I found a work around for an issue that had been haunting me for about 8 months - figured I'd make sure to spread the solution/workaround. 

I believe this is a viable workaround to #460. 

# 🧩 Summary

Fixes inconsistent PSI class resolution between shaded and unshaded IntelliJ packages in FIR Analysis.

# 🐞 Problem

The compiler embeddable distribution shades IntelliJ classes to org.jetbrains.kotlin.com.intellij.*, while the IDE uses unshaded org.jetbrains.kotlin.* packages.
FIR Analysis is expected to always work with the shaded PSI types, but this behavior depends on the root project target:

JVM targets → use shaded PSI (org.jetbrains.kotlin.com.intellij.*)

Non-JVM targets → use unshaded PSI (org.jetbrains.kotlin.*)

This mismatch causes runtime errors such as:

IllegalArgumentException: class org.jetbrains.kotlin.psi.KtCallExpression 
is not a subtype of class org.jetbrains.kotlin.com.intellij.psi.PsiElement

# 🔧 Fix

Inverts the PSI class check to ensure the correct supertype is selected after element resolution, which allows FIR Analysis to correctly identify the active PSI hierarchy in both JVM and non-JVM contexts.

# ✅ Outcome

FIR diagnostics and analysis now work consistently across all target types.

Prevents ClassCastException and similar PSI mismatch issues in the IDE and compiler phases.

If you open an Android target, your FIR diagnostics should work as expected. 

image

# 🧠 Notes

This behavior difference likely originates from how IntelliJ resolves its dependency classpath for non-JVM roots. The inverted check provides a safe, target-agnostic resolution path.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/diagnostics.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/diagnostics.kt
@@ -11,16 +11,30 @@ import org.jetbrains.kotlin.diagnostics.Severity
 import org.jetbrains.kotlin.diagnostics.SourceElementPositioningStrategies
 
 /**
- * The compiler and the IDE use a different version of this class, so use reflection to find the
- * available version.
+ * The compiler and IDE use different PSI classpaths — the compiler’s embeddable distribution
+ * shades IntelliJ packages to `org.jetbrains.kotlin.com.intellij.*`.
+ *
+ * FIR Analysis is expected to always use the shaded PSI classes, but in practice (unfortunately) this depends on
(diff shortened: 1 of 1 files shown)


ZacSweers:
This doesn't affect #460, just IDE behaviors. I have a separate fix for this in #1120 but appreciate the attempt!

PR #1165, opened 2025-10-12
Wasm wasi runtime support

CharlieTap:
Off the back of this [discussion](https://github.com/ZacSweers/metro/discussions/1159), I realised metro supports Wasm and its just the wasi target that was missing. 

I presume this was missed from the config as its a pretty nascent target, logically nothing really differs for metro between `wasmJs` and `wasmWasi` as wasm is a sandboxed environment and I'd expect the same runtime available.

### DIFF
--- a/samples/integration-tests/build.gradle.kts
+++ b/samples/integration-tests/build.gradle.kts
@@ -1,7 +1,5 @@
 // Copyright (C) 2024 Zac Sweers
 // SPDX-License-Identifier: Apache-2.0
-@file:OptIn(ExperimentalWasmDsl::class, ExperimentalKotlinGradlePluginApi::class)
-
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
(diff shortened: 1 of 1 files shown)


runtime/build.gradle.kts
```diff
@@ -64,6 +64,12 @@ kotlin {
     browser {}
   }
 
+  @OptIn(ExperimentalWasmDsl::class)
+  wasmWasi {
+    binaries.executable()
+    nodejs()
```
CharlieTap, line 70: Nodejs is needed here to ensure commonTests run as the test runner is not able to work with solely the wasm binary

ZacSweers:
Can the integration tests be enabled too?

CharlieTap:
It was a little more tricky than just adding wasmWasi as a target, the dependency on ktor (through PlatformUtils) caused it to fail as they haven't added wasmWasi themselves. Since it wasn't a heavy dependency I've swapped it out with a simple kmp impl

ZacSweers:
Thanks!

Issue #1217, opened 2025-10-17
Add support for Kotlin 2.3.20-dev-670 (FirArrayLiteral → FirCollectionLiteral rename)


Lavmee:
### Description

Kotlin 2.3.20-dev-670 introduced a breaking change in FIR where
FirArrayLiteral was renamed to FirCollectionLiteral.

### Alternatives/workarounds
I implemented a small workaround by replacing FirArrayLiteral with FirCall, but some tests are still failing (Dagger Interop and Aggregation(FirHintGenerationWorks)). #1218

ZacSweers:
Should be compatible in the next release, but not planning to document it in the changelog for now

PR #1218, opened 2025-10-17
feat: wip support for kotlin 2.3.20-dev-670.

Lavmee:
#1217


compiler-compat/k2320_dev_670/src/main/kotlin/dev/zacsweers/metro/compiler/compat/k2320_dev_670/CompatContextImpl.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
+import org.jetbrains.kotlin.ir.declarations.IrVariable
+import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.types.IrTypeSystemContext
+import org.jetbrains.kotlin.ir.util.addFakeOverrides as addFakeOverridesNative
+import org.jetbrains.kotlin.name.CallableId
+
+// Ships in 2025.3-EAP
+public class CompatContextImpl : CompatContext {
```
ZacSweers: Since there's no actual API implementation changes, let's just use class delegation to one of the recent 2.3.x impls

Lavmee: Thanks for the feedback! I’ve updated the PR to use class delegation to one of the recent 2.3.x implementations as suggested.

Also, how should we handle the renaming of FirArrayLiteral to FirCollectionLiteral? The workaround I added earlier (replacing it with FirCall) doesn’t seem very reliable.

Some tests are still failing, as I mentioned earlier — most likely related to this change.

ZacSweers: I'll take a look at the tests. Using a looser type isn't the worst solution to this for me, especially as type renames are a bit trickier than just alternative APIs.

ZacSweers:
It doesn't look like all the necessary dependencies for this version are available. Not really sure we can do anything here, the forward compatibility story is very new and I'm not sure n+3 is in scope yet, nor does it look like these kinds of artifacts are available

```
org/jetbrains/kotlin/serialization/js/ModuleKind
java.lang.NoClassDefFoundError: org/jetbrains/kotlin/serialization/js/ModuleKind
	at org.jetbrains.kotlin.test.services.CompilerConfigurationProviderKt.createCompilerConfiguration(CompilerConfigurationProvider.kt:136)
	at org.jetbrains.kotlin.test.services.CompilerConfigurationProviderImpl.createCompilerConfiguration(CompilerConfigurationProvider.kt:107)
	at org.jetbrains.kotlin.test.services.CompilerConfigurationProviderImpl.getCompilerConfiguration(CompilerConfigurationProvider.kt:102)
	at org.jetbrains.kotlin.test.frontend.fir.FirCliFacade.analyze(FirCliFacade.kt:39)
	at org.jetbrains.kotlin.test.frontend.fir.FirCliFacade.analyze(FirCliFacade.kt:24)
	at org.jetbrains.kotlin.test.model.FrontendFacade.transform(Facades.kt:47)
	at org.jetbrains.kotlin.test.model.FrontendFacade.transform(Facades.kt:31)
	at org.jetbrains.kotlin.test.TestStep$FacadeStep.processModule(TestStep.kt:47)
	at org.jetbrains.kotlin.test.TestRunner.processModule(TestRunner.kt:206)
	at org.jetbrains.kotlin.test.TestRunner.hackyProcessModule(TestRunner.kt:197)
	at org.jetbrains.kotlin.test.TestRunner.processModule(TestRunner.kt:141)
	at org.jetbrains.kotlin.test.TestRunner.runTestPipeline(TestRunner.kt:92)
	at org.jetbrains.kotlin.test.TestRunner.runTestImpl(TestRunner.kt:70)
	at org.jetbrains.kotlin.test.TestRunner.runTest(TestRunner.kt:30)
	at org.jetbrains.kotlin.test.TestRunner.runTest$default(TestRunner.kt:28)
	at org.jetbrains.kotlin.test.runners.AbstractKotlinCompilerTest.runTest(AbstractKotlinCompilerTest.kt:115)
	at dev.zacsweers.metro.compiler.FirDumpTestGenerated.testScratch(FirDumpTestGenerated.java:28)
```

That said, the other tests I've tried in :compiler all pass. I'd be fine accepting a scoped down version of this PR with just the change to use `FirCall` (and a comment explaining why) and maybe that's enough for your use cases.

PR #1230, opened 2025-10-19
Use MapBuilder for simple scalar maps

ZacSweers:
Matches how we handle sets


ZacSweers:
...not thrilled with the amount of duplication potential here 🤔. Going to pursue improving that later

PR #1234, opened 2025-10-20
Fix assisted injection non-matching key error

jackwilsdon:
Previously the factory and constructor key lists were incorrectly swapped, causing the error message to be incorrect.

I also added `@Annotated` annotations to the test, as I think it makes it clearer what it is doing.


jackwilsdon:
Oops, forgot to check the legacy tests - thanks for sorting those 😁

PR #1240, opened 2025-10-21
Allow using @ContributesBinding on assisted factories with `contributesAsInject` enabled

kevinguitar:
We have use cases internally to use `@ContributesBinding` on assisted factories for dependency inversion. Like this:

```kotlin
interface Foo {
  val text: String

  interface Factory {
    fun create(text: String): Foo
  }
}

@AssistedInject
class FooImpl(
  @Assisted override val text: String,
) : Foo {

  @ContributesBinding(AppGraph::class)
  @AssistedFactory
  interface Factory : Foo.Factory {
    override fun create(text: String): FooImpl
  }
}
```

And I noticed the new `contributesAsInject` behavior (https://github.com/ZacSweers/metro/pull/1213) throws a FIR error on this use case. I put up a fix by not reporting assisted factories from the `InjectContructorChecker`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Changelog
 - **Enhancement**: Improve code generation around multibinding collection builders and contributors, using more lazy getters in graph code gen.
 - **Enhancement**: Short-circuit empty map providers to `emptyMap()`.
 - **Enhancement**: Support default values for assisted parameter arguments in top-level function injection.
-- **Enhancement**: Allow using `@ContributesBinding` on assisted factories with `contributesAsInject` enabled.
+- **Enhancement**: Allow using `@Contributes*` annotations on assisted factories with `contributesAsInject` enabled.
 - **Fix**: Compute `Optional` instance lazily when requested as a `Provider<Optional<T>>` and the underlying optional is not empty. Only applies to `@BindsOptionalOf` interop.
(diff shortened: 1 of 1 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/InjectConstructorChecker.kt
```diff
@@ -45,6 +46,9 @@ internal object InjectConstructorChecker : FirClassChecker(MppCheckerKind.Common
     val isInjected = classInjectLikeAnnotations.isNotEmpty() || injectedConstructor != null
     if (!isInjected) return
 
+    val isAssistedFactory = declaration.isAnnotatedWithAny(session, classIds.assistedFactoryAnnotations)
+    if (isAssistedFactory) return
```
ZacSweers: I'm a bit confused why this is here, as this class is for validating `@Inject`-like locations but your PR description says you're wanting to allow `@ContributesBinding` on other types? Why is this here?

kevinguitar: With `contributesAsInject` enabled, `injectLikeAnnotations` includes `@ContributesBinding`, and it means assisted factories (from the given use case) are part of the candidates as well

PR #1247, opened 2025-10-21
Add a fallback to testCompilerVersionProvider

kevinguitar:
Gradle task `generateTests` stopped working after https://github.com/ZacSweers/metro/commit/8815ef410bae22bc223a431c3462d3325f638b7b

### DIFF
--- a/compiler-tests/build.gradle.kts
+++ b/compiler-tests/build.gradle.kts
@@ -13,10 +13,9 @@ sourceSets {
   register("generator230")
 }
 
-val testCompilerVersionProvider =
-  providers.gradleProperty("metro.testCompilerVersion").orElse(libs.versions.kotlin)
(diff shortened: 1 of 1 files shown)


compiler-tests/build.gradle.kts
```diff
@@ -13,9 +13,10 @@ sourceSets {
   register("generator230")
 }
 
-val testCompilerVersionProvider = providers.gradleProperty("metro.testCompilerVersion")
+val testCompilerVersionProvider =
+  providers.gradleProperty("metro.testCompilerVersion").orElse(libs.versions.kotlin)
```
ZacSweers: The default actually should be null in our case as its presence results in different code. I think we want to just not use a provider in the Java exec and instead use the realized string or null if absent

PR #1271, opened 2025-10-23
feat: Add reserved keywords from other platforms & add it to validation logic

### LINKED ISSUE #1198: Harden NameAllocator to more platforms
References:

- JVM
  - https://github.com/JetBrains/kotlin/blob/c6f9120e56d438202b75e8d289cbfabfbab14a14/compiler/fir/checkers/checkers.jvm/src/org/jetbrains/kotlin/fir/analysis/jvm/checkers/declaration/FirJvmInvalidAndDangerousCharactersChecker.kt#L4
  - https://github.com/JetBrains/kotlin/blob/c6f9120e56d438202b75e8d289cbfabfbab14a14/compiler/fir/checkers/checkers.jvm/src/org/jetbrains/kotlin/fir/analysis/jvm/FirJvmNamesChecker.kt#L4
- JS
  - https://github.com/JetBrains/kotlin/blob/c6f9120e56d438202b75e8d289cbfabfbab14a14/compiler/fir/checkers/checkers.js/src/org/jetbrains/kotlin/fir/analysis/js/checkers/declaration/FirJsExportDeclarationChecker.kt#L4
    - which uses `validateDeclarationOnConsumableName` to check against a couple special and reserved keywords and makes sure the sanitized name is the same as the original name. Sanitization is driven by the ES 5.1 spec: https://262.ecma-international.org/5.1/#sec-7.6.
- Native
  - https://github.com/JetBrains/kotlin/blob/c6f9120e56d438202b75e8d289cbfabfbab14a14/compiler/fir/checkers/checkers.native/src/org/jetbrains/kotlin/fir/analysis/native/checkers/FirNativeIdentifierChecker.kt#L4

l2hyunwoo:
## Issues

Resolves #1198 

## Solution

- Add reserved keywords from other platforms

```kotlin
/**
 * Cross-platform reserved keywords that must be avoided in generated identifiers.
 *
 * This set contains reserved keywords from JavaScript (ES5/ES6), C99, and Objective-C
 * that are NOT already included in Kotlin's KEYWORDS constant. By pre-allocating these
 * keywords, we ensure generated code compiles successfully on all Kotlin multiplatform
 * targets (JVM, JS, Native).
 *
 * Sources:
 * - JavaScript: ECMAScript 5.1 spec (https://262.ecma-international.org/5.1/#sec-7.6)
 * - C99: ISO/IEC 9899:1999 C Standard
 * - Objective-C: Apple Objective-C Documentation + Clang keywords
 *
 * Many keywords overlap across platforms (e.g., `break`, `case`, `for`), so this set
 * only contains platform-specific keywords not already blocked by Kotlin.
 */
private val CROSS_PLATFORM_RESERVED_KEYWORDS = setOf(
  // ...etc
)
```

- And add it to validation logic

## AI Usage

- Write comments/kdocs (use claude ai)

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExtensionGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrGraphExtensionGenerator.kt
@@ -42,7 +42,7 @@ internal class IrGraphExtensionGenerator(
   private val parentGraph: IrClass,
 ) : IrMetroContext by context {
 
-  private val nameAllocator = NameAllocator(mode = NameAllocator.Mode.COUNT)
+  private val classNameAllocator = NameAllocator(mode = NameAllocator.Mode.COUNT)
   private val generatedClassesCache = mutableMapOf<CacheKey, IrClass>()
 
   private data class CacheKey(val typeKey: IrTypeKey, val parentGraph: ClassId)
@@ -186,7 +186,7 @@ internal class IrGraphExtensionGenerator(
         .buildClass {
           // Ensure a unique name
           name =
-            nameAllocator
+            classNameAllocator
               .newName("${sourceGraph.name.asString().capitalizeUS()}${Symbols.StringNames.IMPL}")
               .asName()
           origin = Origins.GeneratedGraphExtension


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
```diff
@@ -113,7 +151,11 @@ private constructor(
     preallocateKeywords: Boolean = true,
     mode: Mode = Mode.UNDERSCORE,
   ) : this(
-    allocatedNames = if (preallocateKeywords) KEYWORDS.toMutableSet() else mutableSetOf(),
+    allocatedNames = if (preallocateKeywords) {
+      (KEYWORDS + CROSS_PLATFORM_RESERVED_KEYWORDS).toMutableSet()
```
ZacSweers: nit: let's merge CROSS_PLATFORM_RESERVED_KEYWORDS into `KEYWORDS` and just keep that global set

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
```diff
@@ -22,6 +22,44 @@ import kotlin.uuid.Uuid
...
+  // Objective-C-specific keywords (not already in Kotlin KEYWORDS)
+  "id", "nil", "Nil", "YES", "NO", "SEL", "IMP", "BOOL", "instancetype",
+  "required", "optional", "synthesize", "dynamic", "readonly", "readwrite",
+  "assign", "retain", "copy", "nonatomic", "atomic", "strong", "weak",
+)
+
+/**
+ * Dangerous characters that must be replaced in identifiers.
+ */
+private val DANGEROUS_CHARS = setOf('.', ';', '/', '<', '>', '[', ']')
```
ZacSweers: Let's make this just a string like

```suggestion
private const val DANGEROUS_CHARS = ".l/<>[]"
```

The `in` check will still work but it's a bit faster

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
```diff
@@ -180,12 +240,18 @@ internal fun toJavaIdentifier(suggestion: String) = buildString {
...
-        '_'.code
-      }
+    // Determine the valid code point to use
+    val validCodePoint: Int = when {
+      // Explicitly block dangerous characters
+      char in DANGEROUS_CHARS -> '_'.code
+      // Block non-ASCII for cross-platform compatibility (code point > 127)
+      codePoint > 127 -> '_'.code
+      // Use Java identifier validation for other characters
+      Character.isJavaIdentifierPart(codePoint) -> codePoint
```
ZacSweers: Let's do the codePoint checks first before looking up the char. Also, rather than `val char = suggestion[i]` let's just use `codePoint.toChar()`

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
```diff
@@ -168,10 +210,28 @@ private constructor(
...
+ * 3. Replaces other invalid identifier characters with `_`
+ * 4. Prepends `_` if the identifier starts with a digit or other invalid start character
+ *
+ * Examples:
+ * - `foo.bar` → `foo_bar` (dangerous character)
+ * - `café` → `caf_` (non-ASCII)
+ * - `123abc` → `_123abc` (starts with digit)
+ * - `foo<T>` → `foo_T_` (dangerous characters)
+ */
+internal fun toValidatedIdentifier(suggestion: String) = buildString {
```
ZacSweers: nit: let's call this `toSafeIdentifier`

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -624,7 +624,7 @@ internal fun assignConstructorParamsToFields(
       val irField =
         clazz
           .addField(
-            toJavaIdentifier(irParameter.name.asString()),
+            toValidatedIdentifier(irParameter.name.asString()),
```
ZacSweers: note to self: this should probably ask for a `NameAllocator` from somewhere rather than calling this API directly

ZacSweers (commented):
few small nits!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
l2hyunwoo: https://github.com/ZacSweers/metro/pull/1271/commits/cb8faa3153dcbc819ddb7c937749f6a0c7caf258

review applied

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
l2hyunwoo: https://github.com/ZacSweers/metro/pull/1271/commits/225a4eb87176744d9db62b054909bf735e928150 applied 👍🏻

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
l2hyunwoo: https://github.com/ZacSweers/metro/pull/1271/commits/acf7ce92379f069497df11fdc2b2815719ea3b80 applied

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
l2hyunwoo: https://github.com/ZacSweers/metro/pull/1271/commits/914fa80fe58e6215d03863d1e701ac42f9a6c928 applied 👏🏻

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/NameAllocator.kt
```diff
@@ -180,12 +240,18 @@ internal fun toJavaIdentifier(suggestion: String) = buildString {
...
+    // Determine the valid code point to use
+    val validCodePoint: Int = when {
+      // Block non-ASCII for cross-platform compatibility (code point > 127)
+      codePoint > 127 -> '_'.code
+      // Explicitly block dangerous characters
+      codePoint.toChar() in DANGEROUS_CHARS -> '_'.code
+      // Use Java identifier validation for other characters
+      Character.isJavaIdentifierPart(codePoint) -> codePoint
+      // Replace any other invalid character with underscore
+      else -> '_'.code
```
ZacSweers:
```suggestion
    val validCodePoint: Int = when {
      // Block non-ASCII for cross-platform compatibility (code point > 127)
      codePoint > 127 -> '_'.code
      // Use Java identifier validation for other characters
      Character.isJavaIdentifierPart(codePoint) -> codePoint
      // Explicitly block dangerous characters
      codePoint.toChar() in DANGEROUS_CHARS -> '_'.code
      // Replace any other invalid character with underscore
      else -> '_'.code
```

PR #1277, opened 2025-10-24
(WIP) Raise compiler jvmTarget to 21

ZacSweers:


ZacSweers:
Most of these features don't seem fully supported until IJ 2025.3 so gonna pause for now

PR #1301, opened 2025-10-30
Graph Sharding

izantech:
## Graph Sharding

Adds automatic graph sharding for large dependency graphs to avoid JVM method size limits.

### Motivation

Large graphs generate classes that can hit the JVM's 65535-byte method size limit. This splits binding initialization across multiple inner shard classes when graphs exceed a configurable threshold, while keeping provider fields on the root graph class.

### What's included

- **Sharding infrastructure**
  - `IrGraphShardGenerator` - Generates inner shard classes with `initialize()` methods that set up provider bindings
  - `GraphPartitioner` - SCC-aware partitioning algorithm that keeps circular dependencies together
  - `ShardingDiagnostics` - Generates detailed sharding plan reports
  - Preserves circular dependencies by grouping strongly connected components in the same shard

- **Statement chunking**
  - Splits large shard `initialize()` methods into smaller `init1()`, `init2()`, etc. when exceeding `statementsPerInitFun` limit
  - Prevents individual methods from hitting bytecode limits

- **Configuration**
  - `MetroOptions.enableGraphSharding` - Enable/disable sharding (default: **disabled**)
  - `MetroOptions.keysPerGraphShard` - Bindings per shard threshold (default: 2000)

- **Diagnostics**
  - Reports written to `sharding-plan-{GraphName}.txt` showing:
    - Shard distribution and initialization order
    - Cross-shard dependency counts per shard
    - Binding keys per shard (truncated for large shards)
    - Cross-shard dependency mapping (truncated at 100 entries)
  - Warning emitted when sharding is not applied due to custom `keysPerGraphShard` being larger than binding count

- **Testing**
  - Box tests for runtime validation (cycles, provides, binds, multibindings, graph extensions)
  - IR dump tests for generated code structure verification
  - Validation tests for `keysPerGraphShard` option

### Architecture

```kotlin
class AppGraph$Impl : AppGraph {
  // Provider fields for all bindings
  private val repoProvider: Provider<Repository>
  private val apiProvider: Provider<ApiService>
  // ... more providers ...

  // Shard instance fields
  private val shard1: Shard1
  private val shard2: Shard2

  private inner class Shard1 {
    fun initialize(graph: Impl) {
      graph.repoProvider = provider { ... }
      graph.apiProvider = provider { ... }
    }
  }

  private inner class Shard2 {
    // When statements exceed limit, they're chunked:
    private fun init1(graph: Impl) { ... }
    private fun init2(graph: Impl) { ... }

    fun initialize(graph: Impl) {
      init1(graph)
      init2(graph)
    }
  }

  init {
    shard1 = Shard1()
    shard2 = Shard2()
    shard1.initialize(this)
    shard2.initialize(this)
  }
}
```

### Configuration

```kotlin
metro {
  enableGraphSharding = true  // default: false
  keysPerGraphShard = 1200    // tune per project (default: 2000)
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 **Unreleased**
 --------------
 
+- **New**: Experimental support for sharding large graphs. For extremely large dependency graphs on the JVM, their generated implementations could exceed the JVM class size limit. To avoid this, Metro now supports sharding within graphs (as needed) to distribute initialization code across multiple inner _shard_ classes. This is currently disabled by default but can be enabled via the `enableGraphSharding` Gradle DSL property.
+
 0.8.2
 -----
 


izantech:
Hi again, @ZacSweers!

I've finished working on graph sharding in my fork, and I think it's now ready for you to review it. Finally I decided to leave fastInit out of this PR.

I've added some tests (box and IR) but I can add more if you think I missed anything.
Also let me know if you think some more extensive documentation would be appropriate.

Thanks 🙂

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrShardGenerator.kt
```diff
@@ -0,0 +1,137 @@
...
+import org.jetbrains.kotlin.ir.builders.irReturn
+import org.jetbrains.kotlin.ir.builders.irSetField
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.util.addChild
+import org.jetbrains.kotlin.ir.util.copyTo
+import org.jetbrains.kotlin.ir.util.createThisReceiverParameter
+import org.jetbrains.kotlin.ir.util.defaultType
+import org.jetbrains.kotlin.name.Name
+
+internal object IrShardGenerator {
```
ZacSweers: Please document this. How it works, what shape of code it's generating, etc

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingDiagnostics.kt
```diff
@@ -0,0 +1,101 @@
...
+
+import dev.zacsweers.metro.compiler.MetroOptions
+import dev.zacsweers.metro.compiler.ir.IrTypeKey
+import dev.zacsweers.metro.compiler.ir.graph.IrBindingGraph
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.util.kotlinFqName
+
+internal object ShardingDiagnostics {
+
+  /** Generates a diagnostic report showing the sharding plan for a graph. */
```
ZacSweers: This doc does not really help explain what this is returning, could you add an example?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrBindingGraph.kt
```diff
@@ -284,7 +290,7 @@ internal class IrBindingGraph(
           "Found absent bindings in the binding graph: ${dumpGraph("Absent bindings", short = true)}"
         }
       }
-      return BindingGraphResult(sortedKeys, deferredTypes, reachableKeys, false)
+      return BindingGraphResult(sortedKeys, deferredTypes, reachableKeys, false, topologyResult)
```
ZacSweers: We do we pass the keys and the original result they came from here now?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt
```diff
@@ -177,13 +191,17 @@ internal class IrGraphGenerator(
   /**
    * Graph extensions may reserve property names for their linking, so if they've done that we use
    * the precomputed property rather than generate a new one.
+   *
+   * Note: Properties use internal visibility (not private) to allow shard classes to access and
+   * initialize them during sharded graph generation. Shards are nested classes that need to set
+   * these properties in their initialize() methods.
```
ZacSweers: I'm not sure I understand this change. Nested classes can access private members of parents. Also, why nested and not inner (which is what graph extension impls are)

ZacSweers (commented):
Hi @IzanRodrigo. To be honest, this PR is not reviewable/mergeable in its current state due to its sheer size and lack of documentation/descriptions in much it. Please break this up into smaller, readable commits with clear descriptions. It also looks like quite a bit of this is generated with AI, and I'd like to reference to the PR template that asks that this kind of use be disclosed upfront and cleaned up before opening the PR.

I've left a few inline comments from some light reading through

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingDiagnostics.kt
izantech: Yes, here's an example of the report in the android-app sample using this metro config:
```kotlin
metro {
  debug = true
  keysPerGraphShard = 2
  reportsDestination = layout.buildDirectory.dir("metro-reports")
}
```

```
=== Metro Graph Sharding Plan ===

Component: dev.zacsweers.metro.sample.android.AppGraph.$$MetroGraph
Total bindings: 5
Keys per shard limit: 2
Shard count: 3
Sharding enabled: true

Initialization order: Shard1 → Shard2 → Shard3

Shard 1:
  Class: Shard1
  Bindings: 2
  Outgoing cross-shard edges: 0
  Instance property: shard1
  Initialize function: initialize
  Binding keys:
    - android.content.Context
    - dev.zacsweers.metro.sample.android.SampleWorker

Shard 2:
  Class: Shard2
  Bindings: 2
  Outgoing cross-shard edges: 2
  Instance property: shard2
  Initialize function: initialize
  Binding keys:
    - dev.zacsweers.metro.sample.android.SampleWorker.Factory
    - dev.zacsweers.metro.sample.android.SecondWorker

Shard 3:
  Class: Shard3
  Bindings: 1
  Outgoing cross-shard edges: 1
  Instance property: shard3
  Initialize function: initialize
  Binding keys:
    - dev.zacsweers.metro.sample.android.SecondWorker.Factory

Cross-shard dependencies:
  Shard2.dev.zacsweers.metro.sample.android.SampleWorker.Factory → Shard1.dev.zacsweers.metro.sample.android.SampleWorker
  Shard2.dev.zacsweers.metro.sample.android.SecondWorker → Shard1.android.content.Context
  Shard3.dev.zacsweers.metro.sample.android.SecondWorker.Factory → Shard2.dev.zacsweers.metro.sample.android.SecondWorker

Total cross-shard dependencies: 3
```
I used this while developing but maybe it's not that useful now.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt
izantech: I was getting property access errors when the fields were private, but I'll try to use inner class instead.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrShardGenerator.kt
izantech: Okay I will. Should I also create a more detailed document about sharding?

izantech:
> Hi @IzanRodrigo. To be honest, this PR is not reviewable/mergeable in its current state due to its sheer size and lack of documentation/descriptions in much it. Please break this up into smaller, readable commits with clear descriptions. It also looks like quite a bit of this is generated with AI, and I'd like to reference to the PR template that asks that this kind of use be disclosed upfront and cleaned up before opening the PR.
> 
> I've left a few inline comments from some light reading through

Thanks for the feedback, I appreciate it!

I'm going to try to make shards inner classes to see if I fix the property access errors. If it works, most of the changes of this PR (private -> internal) will be reverted. After that I'll create the more readable commits.
About the code generated by AI (most of the ShardingOrchestrator for example), what do you want me to do? I'm okay with adding some disclaimer about it if you want.

I'll get back to you when I address your comments, thank you 🙂

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt
ZacSweers: In general, it's usually a sign that you should ask before proceeding if a change is introducing a large diff to existing tests :)

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrShardGenerator.kt
ZacSweers: Ideally in the kdoc, but also ok to add markdown doc somewhere if that's better

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingDiagnostics.kt
ZacSweers: Please add it to the kdoc (in simplified terms)

ZacSweers:
> About the code generated by AI (most of the ShardingOrchestrator for example), what do you want me to do?

Per the template I want you to de-fluff it and demonstrate that you have independently reviewed and refined it. It's ok if AI is used to assist some work, but AI-generated code on its own is rarely good enough in first pass on a large context project like this. Acknowledging its use alone goes a long way towards demonstrating this (rather than feel like it's been slid by), and generally being mindful on maintainers' time. Especially when it's embedded in a multi-thousand line diff like this.

Basically, I shouldn't be the first human reviewing it and the ask for upfront acknowledgement of that in PRs is there to nudge that.

izantech:
> > About the code generated by AI (most of the ShardingOrchestrator for example), what do you want me to do?
> 
> Per the template I want you to de-fluff it and demonstrate that you have independently reviewed and refined it. It's ok if AI is used to assist some work, but AI-generated code on its own is rarely good enough in first pass on a large context project like this. Acknowledging its use alone goes a long way towards demonstrating this (rather than feel like it's been slid by), and generally being mindful on maintainers' time. Especially when it's embedded in a multi-thousand line diff like this.
> 
> Basically, I shouldn't be the first human reviewing it and the ask for upfront acknowledgement of that in PRs is there to nudge that.

Hello Zac.

I've been working on addressing your previous comments:
- Now the commits are smaller (except the ones where I added tests) and can be reviewed independently.
- I've reviewed all files in the PR to remove any redundant ai-generated code.
- Also I've documented the files you mentioned.

About the property access errors I mentioned, I could revert the visibility changes to private. However, I had to change the visibility of the backing fields to internal to make it work: https://github.com/IzanRodrigo/metro/blob/5213e2b38b16ec256e1955d5019f946725fdac39/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt#L570-L571
Maybe I'm missing something, but I couldn't find a better way to do it 😕

Finally I've been testing this in an app with ~2300 bindings and chained graph extensions and everything seems to be working fine.

When you have the time, please take a look at the PR and let me know if you want me to change/explain anything.
Thank you! 🙂

ZacSweers:
Thanks! I'll try to take a look soon. In the meantime I'll enable CI on this PR in case you see any issues there

gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroPluginExtension.kt
```diff
@@ -115,6 +115,17 @@ constructor(layout: ProjectLayout, objects: ObjectFactory, providers: ProviderFa
   public val statementsPerInitFun: Property<Int> =
     objects.property(Int::class.javaObjectType).convention(25)
 
+  /** Enable/disable graph sharding of binding graphs. Enabled by default. */
+  public val enableGraphSharding: Property<Boolean> =
+    objects.property(Boolean::class.javaObjectType).convention(true)
```
ZacSweers: nit: given the large scope of this, I'm inclined to disable by default to start so we can let more folks try it before making it the default

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+import dev.zacsweers.metro.compiler.graph.TopoSortResult
+import dev.zacsweers.metro.compiler.ir.IrTypeKey
+import dev.zacsweers.metro.compiler.ir.graph.DependencyGraphNode
+import dev.zacsweers.metro.compiler.isInvisibleGeneratedGraph
+
+/**
+ * Partitions bindings into shards when a graph exceeds [MetroOptions.keysPerGraphShard].
+ *
+ * Keeps SCCs together to preserve circular dependencies (including Provider<T> cycles), respects
+ * topological ordering, and merges small tail shards.
```
ZacSweers: what are "small tail shards"?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * Returns null if sharding isn't needed:
+ * - Graph sharding is disabled
+ * - There are too few bindings
+ * - It's either a dynamic graph or a graph extension
```
ZacSweers: Why is this not applicable to dynamic graphs or extensions? There will be a new type soon-ish too - anonymous graphs - that's similar to a dynamic graph but has no source declaration and are instead synthetically generated based on injection targets and input bindings/containers

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/graph/topologicalSort.kt
```diff
@@ -120,11 +120,19 @@ internal fun <TypeKey : Comparable<TypeKey>, Binding> buildFullAdjacency(
  * @param sortedKeys Topologically sorted list of keys.
  * @param deferredTypes Vertices that sit inside breakable cycles.
  * @param reachableKeys Vertices that were deemed reachable by any input roots.
+ * @param adjacency The reachable adjacency map used for topological sorting.
+ * @param components The strongly connected components computed during sorting.
+ * @param componentOf Mapping from vertex to component ID.
+ * @param componentDag The DAG of components (edges between component IDs).
  */
 internal data class TopoSortResult<T>(
```
ZacSweers: Seems this should have a different name now, maybe just `GraphTopology`?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+   * dependencies (including Provider<T> cycles) stay in the same shard. Adds isolated keys at
+   * the end and merges small tail shards.
+   */
+  private inline fun partitionUsingSCCs(
+    containsKey: (IrTypeKey) -> Boolean,
+    sortedKeys: List<IrTypeKey>,
+    components: List<Component<IrTypeKey>>,
+    componentOf: Map<IrTypeKey, Int>,
+    maxBindingsPerShard: Int,
+  ) = buildList {
```
ZacSweers: I find it harder to read the implicit receivers with this pattern, like the `containsKey()` on line 76. Let's extract a local mutable list var for readability

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+    var currentShard = mutableListOf<IrTypeKey>()
+
+    // Build ordered component ids from sortedKeys and collect isolated keys
```
ZacSweers: What are isolated keys in this context?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+internal class ShardingOrchestrator(
+  private val node: DependencyGraphNode,
+  private val options: MetroOptions,
+) {
+
+  /**
+   * Returns shard groups in topologically sorted order, or null if sharding isn't needed.
+   * Each list contains type keys for the bindings in that shard.
+   */
+  fun computeShardGroups(topologyData: TopoSortResult<IrTypeKey>?): List<List<IrTypeKey>>? {
```
ZacSweers: Seems we should just never call this if we have a null result? This is something I see AI do often and I'd like to avoid it since it creates ambiguous signatures (does it return null only if the input is null? etc). At a quick glance I don't see any callers that would pass a nullable input

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+    val keyShards =
+      partitionUsingSCCs(
+        containsKey = topologyData.adjacency::containsKey,
+        sortedKeys = topologyData.sortedKeys,
+        components = topologyData.components,
+        componentOf = topologyData.componentOf,
+        maxBindingsPerShard = maxPerShard,
+      )
+
+    return keyShards.ifEmpty { null }
```
ZacSweers: This seems like it would be an error?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
```diff
@@ -0,0 +1,114 @@
...
+    maxBindingsPerShard: Int,
+  ) = buildList {
+    var currentShard = mutableListOf<IrTypeKey>()
+
+    // Build ordered component ids from sortedKeys and collect isolated keys
+    val orderedComponents = sortedKeys.mapNotNull { componentOf[it] }.distinct()
+    val isolatedKeys = sortedKeys.filter { componentOf[it] == null && containsKey(it) }
+
+    // Append components to shards, respecting size limits
+    for (cid in orderedComponents) {
```
ZacSweers: nit: let's avoid overly terse names like `cid`. We're not writing C code here 🙃

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrGraphShardGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+        instanceProperty = shardInstanceProperty,
+        initializeFunction = initializeFunction,
+        bindings = bindings,
+      )
+    }
+  }
+}
+
+/** Property with its type key and initializer. */
+internal data class PropertyBinding(
```
ZacSweers, line 340: note to self: there is some very similar code in IrGraphGenerator for `propertyInitializers` and `propertiesToTypeKeys` that I have earmarked to use IrAttributes in the future to simplify this

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrGraphShardGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+import org.jetbrains.kotlin.ir.util.addChild
+import org.jetbrains.kotlin.ir.util.copyTo
+import org.jetbrains.kotlin.ir.util.createThisReceiverParameter
+import org.jetbrains.kotlin.ir.util.defaultType
+import org.jetbrains.kotlin.name.Name
+
+/**
+ * Generates IR for shard classes when a graph is partitioned.
+ *
+ * Creates inner shard classes with properties and an initialize() function for their bindings.
```
ZacSweers: Can you add a minimal (pseudo)code snippet to this doc of what it generates?

compiler-tests/src/test/data/box/dependencygraph/sharding/CrossShardDependencies.kt
ZacSweers, line 1: Can you take a pass at adding a simple explanatory doc in each of these tests explaining what conditions they are testing?

compiler-tests/src/test/data/dump/ir/dependencygraph/sharding/MinimalGraphShardingTooSmall.kt
ZacSweers, line 1: Seems we should consider emitting a warning in this scenario (if the keys is customized) to let the user know the sharding value is not sharding?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt
```diff
@@ -578,6 +579,17 @@ internal class IrGraphGenerator(
       if (shardGroups.size > 1) {
         val shardInfos = shardGenerator.generateShards(graphClass, shardGroups)
 
+        writeDiagnostic("sharding-plan-${parentTracer.tag}.txt") {
+          ShardingDiagnostics.generateShardingPlanReport(
```
ZacSweers: Rather than splitting it up, can have the shard generator encapsulate this instead?

ZacSweers:
Some top-level comments

---

Initially you had also included switching providers/fast init. Can you expand on how this unlocks that in the future? As best I could tell, this seems a fairly standalone implementation and I'm not sure where switching providers would fit into this

---

In some scenarios, we extract properties with getters for things we don't want to inline. Namely: multibinding expressions, reused unscoped bindings, and any bindings going into multibinding creators. How do they work in this system?

---

Let's pull 07be647a5bf0423db7d3f62ed975d561b23a9fdb to a separate PR. I have a couple questions about it
  - Why was it necessary for this PR?
  - Wouldn't this fail in the way we're using it since we would have a mismatched number of type parameters and type arguments if we're dropping star args?

---

> However, I had to change the visibility of the backing fields to internal to make it work: [IzanRodrigo/metro@5213e2b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt#L570-L571](https://github.com/IzanRodrigo/metro/blob/5213e2b38b16ec256e1955d5019f946725fdac39/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/IrGraphGenerator.kt#L570-L571)
Maybe I'm missing something, but I couldn't find a better way to do it 😕

You haven't actually pasted the error, what was it?

---

It seems like we should have some basic error tests too, for example a <1 shard size

---

Can we abstract the shared generation bits a bit away from IR and make it testable as a standalone? Similar to how we have non-IR unit tests for topo sorting

---

I couldn't find any tests (but maybe I missed them) demonstrating how this handles valid cycles. I know it's mentioned as supported, but I'd like to see a box and IR dump test to ensure this. Especially if it's something like `A-B-(C-D)-E` type of cycle where it's in the middle of the sort.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
izantech: What I meant by that is for example we have 2100 bindings and it's split in three: ~1000 + ~1000 + ~100
In that case maybe creating a third shard is too much and we can add the bindings to the previous one.

ZacSweers: If we're setting an explicit cap, seems like we should firmly respect it and go over (or emit a warning along the way if it's within a certain threshold if the user wants to consider raising it to reduce shards)

izantech:
> Let's pull [07be647](https://github.com/ZacSweers/metro/commit/07be647a5bf0423db7d3f62ed975d561b23a9fdb) to a separate PR. I have a couple questions about it
> 
> * Why was it necessary for this PR?
> * Wouldn't this fail in the way we're using it since we would have a mismatched number of type parameters and type arguments if we're dropping star args?

I added that commit to fix https://github.com/ZacSweers/metro/issues/1329, but since you fixed it already it's no longer needed and I've removed it 🙂

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
izantech: I suppose we could enable it for every kind of graphs yeah. I thought only main graphs would likely need it but there might be some cases where graph extensions or dynamic graphs are too big.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/ShardingOrchestrator.kt
izantech: Makes sense, I'll change it.

izantech:
Hi there! Thanks for the review, I'll start working on addressing those comments 🙂

About the visibility issues, sorry I forgot to paste the error:
```kotlin
class TestGraph$Impl$Shard1 tried to access private field TestGraph$Impl.service1Provider (TestGraph$Impl$Shard1 and TestGraph$Impl are in unnamed module of loader org.jetbrains.kotlin.codegen.GeneratedClassLoader @6d94f697)
java.lang.IllegalAccessError: class TestGraph$Impl$Shard1 tried to access private field TestGraph$Impl.service1Provider (TestGraph$Impl$Shard1 and TestGraph$Impl are in unnamed module of loader org.jetbrains.kotlin.codegen.GeneratedClassLoader @6d94f697)
	at TestGraph$Impl$Shard1.initialize(MinimalGraphSharding.kt)
	at TestGraph$Impl.<init>(MinimalGraphSharding.kt:12)
	at TestGraph$Impl.<init>(MinimalGraphSharding.kt)
	at TestGraph$Companion.invoke(MinimalGraphSharding.kt:12)
	at MinimalGraphShardingKt.box(MinimalGraphSharding.kt)
```

ZacSweers:
> java.lang.IllegalAccessError: class TestGraph$Impl$Shard1 tried to access private field TestGraph$Impl

This shouldn't be an issue if the shard is an inner class

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/sharding/IrGraphShardGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+          val graphInstanceParam = addValueParameter("graphInstance", graphClass.defaultType)
+
+          body =
+            createIrBuilder(symbol).irBlockBody {
+              for (binding in bindings) {
+                val property =
+                  binding.property.apply {
+                    isVar = true
+                    isLateinit = true
+                  }
```
izantech: @ZacSweers The property access issue was actually happening because the properties at the graph are being initialized in these "initialize" functions. Those are no longer called in the graph constructor, and they can't be val.
Making them lateinit var works, but I'm not sure this is the right approach.
What do you think?

ZacSweers: Lateinit var is not a valid option. Can you please expand a bit more? I don't see why this can't treat properties the same way we do elsewhere with initializers (nonnull, var, not lateinit). It's not valid in source but it is valid in IR and we leverage that

izantech: I can't figure out how to do it with inner classes, I still get the same error :/

I've made a gist with a quick and dirty way to shard the graphs: https://gist.github.com/IzanRodrigo/f383836effd9f2c0a0567c615f489a73

Whenever you can please take a look because I'm probably missing something.

ZacSweers:
Let's stick with the internal visibility for now and I can look at the rest later. Can you cover the remaining things I raised [here](https://github.com/ZacSweers/metro/pull/1301#issuecomment-3492909358)?

It would also be super helpful if you could write up a simple implementation notes doc in the repo for future reference, as this is a huge change.

izantech:
> Let's stick with the internal visibility for now and I can look at the rest later. Can you cover the remaining things I raised [here](https://github.com/ZacSweers/metro/pull/1301#issuecomment-3492909358)?
> 
> It would also be super helpful if you could write up a simple implementation notes doc in the repo for future reference, as this is a huge change.

Okay, I'll keep working on the comments 🙂
Absolutely, will do!

izantech:
Hi there @ZacSweers! Sorry for the delay, I've been busy the last few weeks.
I've been addressing the comments, check it when you have time.

> In some scenarios, we extract properties with getters for things we don't want to inline. Namely: multibinding expressions, reused unscoped bindings, and any bindings going into multibinding creators. How do they work in this system?

Properties are kept at graph level, so those are not affected by my changes.

> Initially you had also included switching providers/fast init. Can you expand on how this unlocks that in the future? As best I could tell, this seems a fairly standalone implementation and I'm not sure where switching providers would fit into this

Yes, I started to work on switching providers, but I decided to focus on graph sharding because it's what's blocking us from using Metro, so I will resume it after this PR is merged.

ZacSweers:
Figured out the property issue - it seems kotlinc doesn't generate accessors for field _setters_ for us in lowering, but does for just getters (which is what the existing code is doing). It does generate it in simple tests though, so I'll look more into it later. In the mean time, did a few touch ups and docs passes

ZacSweers:
> Yes, I started to work on switching providers, but I decided to focus on graph sharding because it's what's blocking us from using Metro, so I will resume it after this PR is merged.

I was more curious for details on how class sharding relates to switching providers?

Merging this now though, thanks a bunch for working on the feedback and the iteration!

ZacSweers:
Talked to the compiler folks and filed an issue for the synthetic setters here: [KT-82977](https://youtrack.jetbrains.com/issue/KT-82977) Jvm IR lowering does not generate synthetic setters for irSetField calls to parents from nested classes

izantech:
> Talked to the compiler folks and filed an issue for the synthetic setters here: [KT-82977](https://youtrack.jetbrains.com/issue/KT-82977) Jvm IR lowering does not generate synthetic setters for irSetField calls to parents from nested classes

Awesome, thanks for finding the cause 🙂

Issue #1303, opened 2025-10-31
"File name too long" error occurs when diagnostic reports are created for factories with a long enough package + name combo


JoelWilcox:
### Description

This error specifically happens when the reports destination has been set and a report is attempted to be created for a provider factory where "provider-factory-$factoryFqName" exceeds the file name length. We run into this in instances where we've generated a binding container from a custom-annotated inner class of another class, but it could also occur with a source file with sufficiently long package + class name combination.

### Self-contained Reproducer

PR coming shortly with a test to cover this and incremental fix, but here's an example that causes the failure:

```
package com.squareup.feature.featureflags

interface FeatureFlag

object RealProductionLoremIpsumFunctionalityFeatureFlagWrapper {
  @ContributesFeatureFlag
  object AllowLoremIpsumFunctionalitySemiLongNameFeatureFlag : FeatureFlag
}
```

```
package anvil.register.featureflags.com.squareup.feature.featureflags

// In this example, this binding container would normally be Anvil- or KSP-generated code based on the above feature flag and a custom contribution annotation
@BindingContainer
@ContributesTo(scope = AppScope::class)
public object RealProductionLoremIpsumFunctionalityFeatureFlagWrapper_AllowLoremIpsumFunctionalitySemiLongNameFeatureFlag {
  @Provides
  @IntoSet
  public fun providesAllowLoremIpsumFunctionalitySemiLongNameFeatureFlag(): FeatureFlag = RealProductionLoremIpsumFunctionalityFeatureFlagWrapper.AllowLoremIpsumFunctionalitySemiLongNameFeatureFlag
}
```

### Metro version

0.7.2

PR #1304, opened 2025-10-31
Encode factory package name in report directory instead of report file name to reduce likelihood of 'file name too long' exceptions

JoelWilcox:
- Fixes https://github.com/ZacSweers/metro/issues/1303
- Also adds a new directive for setting the reports directory in tests using the compiler test infra
- This error is still possible with sufficiently long class names but is much harder to run into now that the provider report prefix and package are never included in the file name. There could be followup improvements still here to prevent it entirely, e.g. by truncating parts of the class name and adding a name hash.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrMetroContext.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrMetroContext.kt
@@ -15,6 +15,7 @@ import java.nio.file.Path
 import kotlin.io.path.appendText
 import kotlin.io.path.createDirectories
 import kotlin.io.path.createFile
+import kotlin.io.path.createParentDirectories
 import kotlin.io.path.deleteIfExists
 import kotlin.io.path.writeText
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
@@ -261,7 +262,7 @@ internal fun writeDiagnostic(fileName: () -> String, text: () -> String) {
     ?.resolve(fileName())
     ?.apply {
       // Ensure that the path leading up to the file has been created
-      File(this.toString()).mkdirs()
+      createParentDirectories()
       deleteIfExists()
     }
     ?.writeText(text())


compiler-tests/src/test/kotlin/dev/zacsweers/metro/compiler/MetroExtensionRegistrarConfigurator.kt
```diff
@@ -81,6 +83,10 @@ class MetroExtensionRegistrarConfigurator(testServices: TestServices) :
         module.directives.singleOrZeroValue(MetroDirectives.MAX_IR_ERRORS_COUNT)?.let {
           maxIrErrorsCount = it
         }
+        module.directives.singleOrZeroValue(MetroDirectives.REPORTS_DESTINATION)?.let {
```
ZacSweers, line 86: nice

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/BindingContainerTransformer.kt
```diff
@@ -371,9 +372,14 @@ internal class BindingContainerTransformer(context: IrMetroContext) : IrMetroCon
...
-      factoryCls.dumpKotlinLike()
-    }
+    val factoryPath =
+      factoryCls.packageFqName?.let { packageName ->
+        val fileName = factoryCls.kotlinFqName.toString().replace("$packageName.", "")
+        "$packageName/$fileName"
+      } ?: factoryCls.kotlinFqName.asString()
+
+    // Relative path example: provider-factories/dev.zac.feature/Outer.Inner$$Factory.kt
+    writeDiagnostic("provider-factories/$factoryPath.kt") { factoryCls.dumpKotlinLike() }
```
ZacSweers, line 382: What about just making the package structure match the directory structure? i.e. `com/example/Foo.Bar` yields dir `com/example` with `Foo$Bar.kt` (and possible `Foo.kt`) in it?

JoelWilcox: Sounds good to me! Initially left it like this to keep the report dirs in a flatter structure for convenience when looking through them

compiler-tests/src/test/data/box/reports/ReallyLongPackageNameHandledWhenWritingDebugReports.kt
```diff
@@ -0,0 +1,62 @@
...
+ * java.nio.file.FileSystemException: /private/var/folders/wv/bkgjhyrj3l5f8pwkh_kfds000000gn/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Reports2testReallyLongPackageNameHandledWhenWritingDebugReports/metro/reports/provider-factory-anvil.register.featureflags.com.squareup.feature.featureflags.RealProductionLoremIpsumFunctionalityFeatureFlagWrapper_AllowLoremIpsumFunctionalitySemiLongNameFeatureFlag.ProvidesAllowLoremIpsumFunctionalitySemiLongNameFeatureFlag$$MetroFactory.kt: File name too long
+ * 	< ... java internals stacktrace ... >
+ * 	at dev.zacsweers.metro.compiler.ir.IrMetroContextKt.writeDiagnostic(IrMetroContext.kt:265)
+ *
+ * 	In this example, the default report prefix and package name add up to almost 80 characters:
+ * 	provider-factory-anvil.register.featureflags.com.squareup.feature.featureflags
+ *
+ * 	This isn't exclusively a package issue, because it's still a combination of the related factory class name, but encoding the package as a report path segment instead of the file name is a simple way to significantly limit the cases where this happens without reducing generated code or file readability.
+ */
+// REPORTS_DESTINATION: metro/reports
```
ZacSweers, line 13: Would be cool if we could expose something in the test infra/base class to make this easy to grab

PR #1313, opened 2025-11-03
Failing test for member injecting a parent's abstract val

kevinguitar:
We encountered a crash on 0.7.3 while member-injecting a class that has an abstract val in its parent.

This test failed with:
```
class kotlin.collections.builders.SetBuilder cannot be cast to class java.lang.String (kotlin.collections.builders.SetBuilder is in unnamed module of loader java.net.URLClassLoader @66be3cd0; java.lang.String is in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class kotlin.collections.builders.SetBuilder cannot be cast to class java.lang.String (kotlin.collections.builders.SetBuilder is in unnamed module of loader java.net.URLClassLoader @66be3cd0; java.lang.String is in module java.base of loader 'bootstrap')
	at MyClass$MetroMembersInjector.injectMembers(module_lib_InjectingParentAbstractVal.kt:8)
	at MyClass$MetroMembersInjector.injectMembers(module_lib_InjectingParentAbstractVal.kt:8)
	at Module_main_InjectingParentAbstractValKt.box(module_main_InjectingParentAbstractVal.kt:29)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at org.jetbrains.kotlin.test.backend.handlers.JvmBoxRunner.runBoxInCurrentProcess(JvmBoxRunner.kt:169)
	at org.jetbrains.kotlin.test.backend.handlers.JvmBoxRunner.callBoxMethodAndCheckResult(JvmBoxRunner.kt:140)
	at org.jetbrains.kotlin.test.backend.handlers.JvmBoxRunner.callBoxMethodAndCheckResultWithCleanup(JvmBoxRunner.kt:91)
```

Few observations:
- The test passed on 0.7.2
- Without modularization, the test passed fine

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/MembersInjectorTransformer.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/MembersInjectorTransformer.kt
@@ -494,8 +494,42 @@ internal class MembersInjectorTransformer(context: IrMetroContext) : IrMetroCont
 
     val companionObject = injectorClass.companionObject() ?: return emptyList()
 
-    return injectFunctionNames.mapNotNull { functionName ->
-      // Find the inject function by name
+    // Try to get create() function to determine the correct parameter order
+    val createFunction = companionObject.requireSimpleFunction(Symbols.StringNames.CREATE).owner
+
+    val allCreateParams = createFunction.regularParameters
+
+    // Match each inject function to its position in create() params by parameter name
+    data class MatchedFunction(val functionName: String, val startPosition: Int)
+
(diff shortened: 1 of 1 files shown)


ZacSweers:
I've got a so-so solution for this for now, but a better change would be to fully split the sourcing here to use create() functions for metro injectors and read from sources directly for dagger injectors. Since that's a bigger change, I'm going to punt that for now. In the mean time, this solution has one edge case with overloaded injected setters, but that's an edge case I'm willing to deal with until I can do the better change.

Issue #1325, opened 2025-11-04
Provider return type cannot be resolved when defined in the same binding container


JoelWilcox:
### Description

When a binding container defines a type A and also contains an `@Provides` method with return type A, the return type cannot be resolved and results in an error like:
```
Somewhere in file /private/var/folders/wv/bkgjhyrj3l5f8pwkh_kfds000000gn/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Bindingcontainers7641testProviderForTypeDefinedInContainer/kotlin-sources/main/ProviderForTypeDefinedInContainer.kt: java.lang.IllegalStateException:                 Could not resolve provider return type for provider: /MyContainer.provideMyType
                This can happen if the provider references a class that is nested within the same parent class and has cyclical references to other classes.
                @R|dev/zacsweers/metro/Provides|() public? final? fun provideMyType(): MyType {
    ^provideMyType object : MyType {
        private constructor(): R|<anonymous>| {
            super<<implicit>>()
        }

    }

}
. This is a bug in the Metro compiler, please report it to https://github.com/zacsweers/metro. 
	at org.jetbrains.kotlin.util.AnalysisExceptionsKt.wrapIntoFileAnalysisExceptionIfNeeded(AnalysisExceptions.kt:62)
	at org.jetbrains.kotlin.fir.FirCliExceptionHandler.handleExceptionOnFileAnalysis(Utils.kt:267)
	at org.jetbrains.kotlin.fir.resolve.transformers.FirSupertypeResolverTransformer.transformFile(FirSupertypesResolution.kt:861)
```

### Self-contained Reproducer

This is a minimal reproducer:
```
@BidningContainer
@ContributesTo(AppScope::class)
object MyContainer {
  @Provides fun provideMyType(): MyType {
    return object : MyType
  }

  interface MyType
}

@SingleIn(AppScope::class)
@DependencyGraph(AppScope::class)
interface AppGraph {
  val myType: MyContainer.MyType
}
```

### Metro version

0.7.4

ZacSweers:
Marking this as "waiting for upstream" because we can't really do this unless we can move provider factory class generation solely to to IR (https://youtrack.jetbrains.com/issue/KT-79565)

JoelWilcox:
Makes sense! Any preference on me leaving the draft PR open or closing until the upstream ticket is resolved?

ZacSweers:
We can close it for now and revisit once we can do something concrete

PR #1326, opened 2025-11-04
Handle providers that return nested types contained in the same binding container

JoelWilcox:
- Add a failing test for https://github.com/ZacSweers/metro/issues/1325
- WIP: Try out using MetroFirTypeResolver to fix the binding container embedded type provider issue. Currently this results in a "Symbol not found" error compared to the original FIR error type error.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
@@ -3,14 +3,18 @@
 package dev.zacsweers.metro.compiler.fir
 
 import org.jetbrains.kotlin.fir.FirSession
+import org.jetbrains.kotlin.fir.declarations.DirectDeclarationsAccess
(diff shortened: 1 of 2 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
```diff
@@ -47,7 +52,12 @@ internal sealed interface MetroFirTypeResolver {
         } ?: return null
       return resolversByFile.getOrPut(file) {
         val scopes = createImportingScopes(file, session, scopeSession)
-        val configuration = TypeResolutionConfiguration(scopes, emptyList(), useSiteFile = file)
+        val containedTypes =
```
JoelWilcox, line 55: This change didn't end up having an effect, but left in for now to show things that have been tried. The hope here was to ensure the nested class is visible to the type resolver but it doesn't seem like that's how the param is actually used

JoelWilcox:
Closing for now because some upstream changes will be required before it can be supported. (https://github.com/ZacSweers/metro/issues/1325)

Issue #1348, opened 2025-11-11
[IC] Restoring a previous multibinding contribution is not detected


jonamireh:
### Description

It appears if a multibinding is contributed via a Dagger Module / Metro BindingContainer, a removal and then restoration in an other compilation does not restore the functionality

### Self-contained Reproducer

https://github.com/ZacSweers/metro/pull/1349

### Metro version

0.7.5

### Context

The actual usage in our project is an annotation linked with a KSP processor will generate a Multibinding (for an Android ViewModel) as a Module/BindingContainer contribution. Removing that annotation removes the generated file because KSP cached the inputs & outputs, restoring the annotation restores the files but it seemed like Metro didn't pick up that restoration

ZacSweers:
Fixed by #1349

PR #1349, opened 2025-11-11
[IC] Fix restoring multibinding contribution not detected

jonamireh:
## Summary
Added a failing test case for #1348


ZacSweers:
Some early investigation: looks like both do compile, need to see why the contribution is missing

CleanShot 2025-11-11 at 15 42 36@2x

ZacSweers:
Another update: Something strange is happening with IC here. While we _do_ get the declaration and look it up via hints, we are _not_ getting its annotations

```
ClassId: test.AppModule
Class:
  CLASS IR_EXTERNAL_DECLARATION_STUB CLASS name:AppModule modality:FINAL visibility:public superTypes:[kotlin.Any]
```

I'm not quite sure what to do about this yet. We do a few things to tick IC but I've never encountered a situation where we're getting lossy symbols (it's usually just no symbols at all)

PR #1354, opened 2025-11-11
Check if Multibinds map keys are enums

### LINKED ISSUE #1353: Multibinds with Enum key causes an error
### Description

If you declare a `Map` with `@Multibinds` that has uses an `enum class` as a key, Metro throws an exception:

```
Multibinding map keys must be a primitive, String, KClass, or an annotation class.
```

Per the docs on `@MapKey`, `enum class` is allowed as a key.

### Self-contained Reproducer

```kotlin
enum class Animal {
    DOG, CAT;
}

@MapKey
annotation class EnumMapKey(val value: Animal)

@DependencyGraph
interface ExampleGraph {
    @Provides
    @IntoMap
    @EnumMapKey(Animal.DOG)
    fun provideDog(): String = "dog"

    @Provides
    @IntoMap
    @EnumMapKey(Animal.CAT)
    fun provideCat(): String = "cat"

    @Multibinds(allowEmpty = true)
    fun primeMap(): Map<Animal, String>

    val directMap: Map<Animal, String>
}

fun box(): String {
    val graph = createGraph<ExampleGraph>()

    assertEquals(mapOf(Animal.DOG to "dog", Animal.CAT to "cat"), graph.directMap)
    return "OK"
}
```

### Metro version

0.7.5

### Context

This only happens if you declare the `Map` with at `@Multibinds`; if you just declare an accessor without it, this works fine so it's isolated to the diagnostic for `@Multibinds`

jonamireh:
## Summary
This PR fixes #1353 by adding a check for map keys being enums in `MultibindsChecker` as [per the docs](https://github.com/ZacSweers/metro/blob/main/runtime/src/commonMain/kotlin/dev/zacsweers/metro/MapKey.kt#L27-L47), they are allowed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 **Unreleased**
 --------------
 
+- **Fix**: Fix diagnostic when an `@Multibinds`-annotated Map uses an enum as a key.
+
 0.7.5
 -----
 


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/checkers/MultibindsChecker.kt
```diff
@@ -127,7 +130,7 @@ internal object MultibindsChecker : FirCallableDeclarationChecker(MppCheckerKind
             } else {
               // Keys can only be const-able or annotation classes
               keyTypeArg.type?.let { keyType ->
-                if (keyType.isPrimitive || keyType.isString || keyType.isKClassType()) {
+                if (keyType.isPrimitive || keyType.isString || keyType.isKClassType() || keyType.toRegularClassSymbol(session)?.isEnumClass == true) {
```
jonamireh: The public `ConeKotlinType.isEnum` returns `false` for actual enums (???) but the internal one which uses the `session` correctly returns `true` so I just copied the implementation. Let me know if this should be lifted somewhere else or otherwise documented

ZacSweers (approved):

Issue #1366, opened 2025-11-14
A duplicate binding error is reported for a manually-constructor-injected Dagger Module declared in a Component annotation [Dagger Interop]


JoelWilcox:
### Description

When you have a Dagger `@Module` that you'll be instantiating manually with one or more constructor args, in Dagger you can just declare it in the `@Component.modules` field or declare it in that field AND specify it as a `@Component.Factory.create` param. But Dagger will give you an error if you make it a factory param without declaring it in the included modules. Metro takes kind of the opposite stance where both of Dagger's expected scenarios are errors, and it must only be declared in the factory. This use-case results in an error like
```
java.lang.IllegalStateException: //private/var/folders/wv/bkgjhyrj3l5f8pwkh_kfds000000gn/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Interop$Dagger2testTestScratch/kotlin-sources/main/TestScratch.kt:(425,437): error: [Metro/DuplicateBinding] Multiple bindings found for ProvidersModule

  file:///private/var/folders/wv/bkgjhyrj3l5f8pwkh_kfds000000gn/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Interop$Dagger2testTestScratch/kotlin-sources/main/TestScratch.kt:27:7
    BoundInstance(ProvidersModule)
  file:///private/var/folders/wv/bkgjhyrj3l5f8pwkh_kfds000000gn/T/dev.zacsweers.metro.compiler.BoxTestGenerated$Interop$Dagger2testTestScratch/kotlin-sources/main/TestScratch.kt:11:1
    BoundInstance(ProvidersModule)
```
when I would be expecting to see the same error shown for binding containers:
```
java.lang.IllegalStateException: BINDING_CONTAINER_ERROR: Included binding container 'ProvidersModule's does not have a no-arg constructor and thus cannot be included via annotation. Add a no-arg constructor or declare it as a graph factory parameter instead
```

### Self-contained Reproducer

```
@Module
class ProvidersModule(val manuallyProvidedName: String) {
  @Provides
  fun provideString(): String = "Hello $manuallyProvidedName"
}

@Singleton
@Component(
  modules = ProvidersModule::class,
)
interface SingletonComponent {
  val string: String

  @Component.Factory
  interface Factory {
    fun create(
      @Includes providersModule: ProvidersModule,
    ): SingletonComponent
  }
}
```

### Metro version

0.7.5

PR #1367, opened 2025-11-14
Fix Dagger interop error messages when validating `@Module`s with constructor-injected fields to match errors for `@BindingContainer`s

JoelWilcox:
- Fixes https://github.com/ZacSweers/metro/issues/1366


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
```diff
@@ -874,6 +874,12 @@ internal fun FirAnnotation.additionalScopesArgument() =
 internal fun FirAnnotation.bindingContainersArgument() =
   arrayArgument(Symbols.Names.bindingContainers, index = 4)
 
+internal fun FirAnnotation.modulesArgument() = arrayArgument(Symbols.Names.modules, index = 1)
+
+internal fun FirAnnotation.bindingContainerClasses(includeModulesArg: Boolean): FirCall? {
+  return bindingContainersArgument() ?: if (includeModulesArg) modulesArgument() else null
+}
```
JoelWilcox, line 881: This mirrors how we're handling the interop in IR

ZacSweers (approved):

PR #1401, opened 2025-11-22
WIP Kotlin 2.3.0 final

ZacSweers:


ZacSweers:
Abandoning this for now, will revisit when we want to drop 2.2.2x

PR #1426, opened 2025-11-26
Introduce metrox-viewmodel artifacts

ZacSweers:
Changed a bit from the samples. I didn't really see a compelling use case for the whole `ViewModelGraphProvider` setup and custom local graphs, it seemed like abstraction without much value prop. This instead focuses on providing foundational plumbing for the common cases.

Added a generic to the manual injection case to let the compiler help enforce alignment, but I could be convinced to remove this to streamline not needing to add a `binding<ManualViewModelAssistedFactory<*>>()` every time.

One thing not super clear to me is what the KMP story is for viewmodels. Both androidx and JB seem to claim to have KMP artifacts, but maybe only JB has a KMP `viewModel()` API for compose?

https://developer.android.com/kotlin/multiplatform/viewmodel
https://kotlinlang.org/docs/multiplatform/compose-viewmodel.html#using-viewmodel-in-common-code


samples/compose-viewmodels/core/src/commonMain/kotlin/dev/zacsweers/metro/sample/composeviewmodels/core/ViewModelScope.kt
```diff
@@ -1,5 +0,0 @@
-// Copyright (C) 2025 Zac Sweers
-// SPDX-License-Identifier: Apache-2.0
-package dev.zacsweers.metro.sample.composeviewmodels.core
-
-abstract class ViewModelScope private constructor()
```
yschimke, line 5: Since this came up in Slack, an example of how I use ViewModelScoped in Hilt.

https://github.com/google/horologist/blob/4b39b417003657414a6e3c25a8856f519b13646d/ai/sample/wear-gemini/src/main/java/com/google/android/horologist/ai/sample/wear/gemini/di/ViewModelModule.kt#L28

```
    @ViewModelScoped
    @Provides
    fun selfService(
        @ApplicationContext context: Context,
        lifecycle: ViewModelLifecycle,
    ): InferenceServiceGrpcKt.InferenceServiceCoroutineStub {
        return lookupInferenceService(
            context,
            context.packageName,
        ).apply {
            lifecycle.addOnClearedListener {
                (channel as? ManagedChannel?)?.shutdownNow()
            }
        }
    }
```

So it's two things, one having a defined lifecycle for dependencies in the DI graph. Having a synthetic parameter `lifecycle: ViewModelLifecycle`.

Without this I could do either

1) Just accept that `InferenceService` will be singleton scoped
2) I introduce a Factory and create per instance, and clear manually with the ViewModel.onCleared

metrox-viewmodel/src/commonMain/kotlin/dev/zacsweers/metrox/viewmodel/ViewModelScope.kt
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) 2025 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metrox.viewmodel
+
+/** Scope marker for ViewModel-scoped dependency graphs. */
+public abstract class ViewModelScope private constructor()
```
yschimke, line 6: Doesn't appear to be used?

samples/compose-viewmodels/screen-details/src/commonMain/kotlin/dev/zacsweers/metro/sample/composeviewmodels/details/DetailsViewModel.kt
yschimke, line 20: is it worth adding a non-Assisted dependency, to show how to mix?

hrach:
One of the most dangerous runtime issues is changing a non-assisted VM to an assisted one.
There is no compile-time check to change the call on the compose side.

Have you thought about whether there would be (in the future) something to do about it on the Metro side?

For now, I have added only assisted VM support to my codebase to avoid such issues. 

----

Do I understand correctly that SavedStateHandle has to be manually read from creation extras in the assisted factory? I loved the original PR that made this a non-assisted dep. (Basically, a majority of all VMs should use this). Also, at first glance, I don't know how I would obtain the SavedStateHandle in the manual injection case. 

----

Since Kotlin misses default generic types (or associated types, anyway, nothing is planned for, afaik), it is quite cumbersome to repeat the VM type... I [came up](https://github.com/ZacSweers/metro/pull/886#issuecomment-3185820262) with "with"-block lambda, but I guess that's too "wild" for a public lib API, right?

```kotlin
val vm = metroViewModel { with<MyViewModel.Factory>().create(1) } 
```

ZacSweers:
> One of the most dangerous runtime issues is changing a non-assisted VM to an assisted one.
> There is no compile-time check to change the call on the compose side.
> 
> Have you thought about whether there would be (in the future) something to do about it on the Metro side?

Sorry I don't really follow, can you expound?

---

> Do I understand correctly that SavedStateHandle has to be manually read from creation extras in the assisted factory?

Yes. It's not my experience that "a majority of all VMs should use this". AndroidX's own docs even deprecated their [base factory for this](https://developer.android.com/reference/androidx/lifecycle/AbstractSavedStateViewModelFactory) in favor of `CreationExtras.createSavedStateHandle`. I think you may be a bit out of date? 

> Deprecated: Use [viewModelFactory](https://developer.android.com/reference/androidx/lifecycle/viewmodel/package-summary#viewModelFactory(kotlin.Function1)) or implement [ViewModelProvider.Factory](https://developer.android.com/reference/androidx/lifecycle/ViewModelProvider.Factory) directly, combined with [CreationExtras.createSavedStateHandle](https://developer.android.com/reference/androidx/lifecycle/package-summary#(androidx.lifecycle.viewmodel.CreationExtras).createSavedStateHandle()). This base class creates a [SavedStateHandle](https://developer.android.com/reference/androidx/lifecycle/SavedStateHandle) for every [ViewModel](https://developer.android.com/reference/androidx/lifecycle/ViewModel), even when unnecessary, causing overhead.

---

> Since Kotlin misses default generic types (or associated types, anyway, nothing is planned for, afaik), it is quite cumbersome to repeat the VM type... I https://github.com/ZacSweers/metro/pull/886#issuecomment-3185820262 with "with"-block lambda, but I guess that's too "wild" for a public lib API, right?

I'm not really sure what you're referring to with this and it would help if you could clarify

samples/compose-viewmodels/screen-details/src/commonMain/kotlin/dev/zacsweers/metro/sample/composeviewmodels/details/DetailsViewModel.kt
ZacSweers: as in just a regular dependency from the graph?

metrox-viewmodel/src/commonMain/kotlin/dev/zacsweers/metrox/viewmodel/ViewModelScope.kt
ZacSweers: Here for common use but not used in any samples, no. This belongs to a graph

samples/compose-viewmodels/screen-details/src/commonMain/kotlin/dev/zacsweers/metro/sample/composeviewmodels/details/DetailsViewModel.kt
yschimke: Yep. But not essential, this is just if it shows a typical use case

hrach:
Ad 1/

My very usual problem is that I need to convert a non-assisted VM to an assisted one. But I forgot to update the use-site to use a different function and pass the arguments. Usually, because when I tackle VM's logic, I don't bother with VM's DIC setup/usage.

E.g., changing the sample's code to use non-assisted `metroViewModel()` results in a runtime crash:
```
java.lang.IllegalArgumentException: Unknown model class class dev.zacsweers.metro.sample.composeviewmodels.details.DetailsViewModel (Kotlin reflection is not available)
                                                                                                    	at dev.zacsweers.metrox.viewmodel.MetroViewModelFactory.create(MetroViewModelFactory.kt:91)
                                                                                                    	at androidx.lifecycle.viewmodel.internal.ViewModelProviderImpl_androidKt.createViewModel(ViewModelProviderImpl.android.kt:35)
```

As I wrote, in my current implementation, I tried to avoid this problem by introducing only a single type of (assisted) instantiation. Also, I opened [this discussion](https://github.com/ZacSweers/metro/discussions/415) exactly because of this. 

The question was: Have you thought about whether there would be (in the future) something to do about it on the Metro side?

------

Ad 2/

I think this is a bit of a misunderstanding. First, SSH should be really needed quite often - yes, the reality is that the majority of Android devs are not tackling process death (config change is "solved" by retaining the VM, but PD needs SSH).

Second, this is not about how to obtain SSH, but how to expose it to the dev's surfacing API. My point was that SSH could be a non-assisted dependency, as the original PR did: 

- https://github.com/yschimke/metro/blob/bae29759edbc8cb0521b68be6e743f008fec953b/samples/compose-navigation-app/src/main/kotlin/dev/zacsweers/metro/sample/androidviewmodel/viewmodel/ViewModelGraph.kt#L27

The point about AbstractSavedStateViewModelFactory is irrelevant; yshimke's version is not using it.

So, adding SSH won't compile now without changing the constructor to assisted injection - suddenly we are in the situation as described in 1/ - it will compile but crash in runtime if I forgot to change the use-site.

------

Ad 3/

I'm just too lazy to type DetailsViewModel twice, as shown in the sample:

```kotlin
assistedMetroViewModel<DetailsViewModel, DetailsViewModel.Factory> { create(data) },
```

Therefore, I came up with 

```kotlin
assistedMetroViewModel { with<DetailsViewModel.Factory>().create(data) },
```

because there is nothing like

```kotlin
fun <VM, VMF = VM.Factory> assistedMetroViewModel(block: VMF.() -> VM): VM
```

The reason I am mentioning this is partially connected (once again) to 1/ - having some fun call intrinsic that would compile-time check that with's type exists...

ZacSweers:
> The question was: Have you thought about whether there would be (in the future) something to do about it on the Metro side?

I answered this already: https://github.com/ZacSweers/metro/discussions/415#discussioncomment-13081052

In general metro is not likely to build more viewmodel extension APIs as I don't think modern apps should be built with them as anything other than plumbing.

> So, adding SSH won't compile now without changing the constructor to assisted injection - suddenly we are in the situation as described in 1/ - it will compile but crash in runtime if I forgot to change the use-site.

Sorry, the android docs recommend just retrieving it from CreationExtras. This is matching that. Not trying to promote it back into top-level API for the same reasons they deprecated it. If you want it in your own top-level API, you don't have to use this artifact and can do so yourself :)

re: 3 - that just seems like code golf tbh. I may remove the generic type, but not in favor of a golf solution

samples/compose-viewmodels/core/src/commonMain/kotlin/dev/zacsweers/metro/sample/composeviewmodels/core/ViewModelScope.kt
ZacSweers: I think in this case it's reasonable that people that need this pattern can implement it on their own

metrox-viewmodel/src/commonMain/kotlin/dev/zacsweers/metrox/viewmodel/ViewModelScope.kt
yschimke: The readme references it, but doesn't work as such.

PR #1437, opened 2025-11-29
Remove replace warning

ZacSweers:


ZacSweers:
Huh, ok so it does propagate

PR #1448, opened 2025-11-30
Backport some work from 2.3.20 testing

ZacSweers:
Ref #1445


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/ContributionHintIrTransformer.kt
```diff
@@ -20,6 +23,14 @@ internal class ContributionHintIrTransformer(
   private val hintGenerator: HintGenerator,
 ) : IrMetroContext by context {
 
+  // Only executed if generateJvmContributionHintsInFir is enabled
```
ZacSweers:
```suggestion
  // Only executed if generateContributionHintsInFir is enabled
```

PR #1474, opened 2025-12-04
Add usage information to AppScope's javadoc

jingibus:
Does what it says on the tin. 

I was looking to migrate from our own AppScope, and wasn't sure if Metro gave it some other meaning from ours automatically (it doesn't).

I might keep ours just to have our own Javadoc, but I thought some clarity here would be nice.


ZacSweers (approved):

ZacSweers:
Added a couple examples too. Thanks!

PR #1482, opened 2025-12-06
Add an edit and view action on the docs pages

ansman:
This will make it easier to submit fixes for the docs.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -43,6 +43,8 @@ theme:
   features:
     - content.code.copy
     - content.code.select
+    - content.action.edit
+    - content.action.view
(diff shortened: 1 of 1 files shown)


hossain-khan:
Nice! I already use it on our docsite! ^_^

hossain-khan:
As I suspected (based on past struggle with `mike` 😅), wanted to double check if things are working 🧐


Screenshot 2025-12-06 at 5 49 47 PM

Screenshot 2025-12-06 at 5 50 04 PM


I think it's half way there, I can try to look into this 404 issue.

ansman:
Fixed here: https://github.com/ZacSweers/metro/pull/1484

PR #1484, opened 2025-12-06
Fix viewing and editing docs pages

ansman:
The default assumes the branch main branch is called "master".


hossain-khan:
👋🏽 I opened a PR fixing this already! :-) 
* https://github.com/ZacSweers/metro/pull/1483

PR #1530, opened 2025-12-13
Add remaining ViewmodelKeys annotation targets

jonhan:
Add remaining annotation targets for all ViewmodelKeys to match those in other MapKeys


ZacSweers:
Can you add a line to the changelog?

PR #1539, opened 2025-12-15
Don't apply replaces effect from excluded contributions

mateherber-jt:
I've hit an issue (or at least I think) when I migrated our codebase to metro: When I exclude a `BindingContainer` that has a replaces annotation, I expect the replaces not having effect, but that is not the case.

With the help of AI I tried to solve this problem, the changes for me looks reasonable. I also did a bit of refactor by replacing sequence and collection operations with standard for loops to avoid allocation as much as possible.

Also added a test case that showcases the issue and which is solved by this change. If I missed something, or I interpret the expected behavior incorrectly, please close the PR!

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributionMerger.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributionMerger.kt
@@ -153,31 +153,39 @@ internal class IrContributionMerger(
       }
     }
 
-    // Process replacements from both regular contributions and binding containers
+    // Process replacements from both regular contributions and binding containers.
+    // Iterate over the mutable collections (post-exclusion) to avoid processing
+    // excluded items.
+    val classesToReplace = mutableSetOf<ClassId>()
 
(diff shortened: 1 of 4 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributionMerger.kt
ZacSweers, line 1: We lost some of the inline comments here with your AI's refactor, can we add inline comments back where appropriate?

ZacSweers:
Can you also add a line to the CHANGELOG?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributionMerger.kt
mateherber-jt: Reintroduced the comments

ZacSweers:
Please just push new commits after review's started rather than force-push every time, as it blows away history and makes it harder to review/follow the changes

PR #1541, opened 2025-12-15
Add a test case for an IC issue

ansman:
Adding an internal contributed interfaces won't be picked up by the next compilation.


ZacSweers:
Talked offline, this is not a supported use case

PR #1551, opened 2025-12-16
Avoid deeply nested IR calls

vRallev:
Metro used the builder pattern to generate the the multibinding set. With large contributed graphs, this could lead to a Stackoverflow since each chained call causes a recursive call in the Kotlin compiler. With this change the builder calls aren't chained anymore to break this cycle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Changelog
 --------------
 
 - **Fix**: Don't apply `replaces` effect from excluded contributions in cross-module scenarios.
-- **Fix**: Avoid deeply nested IR calls for generated Set and Map multibindings causing `StackoverflowErrors` in large graphs.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changelog
 --------------
 
 - **Fix**: Don't apply `replaces` effect from excluded contributions in cross-module scenarios.
+- **Fix**: Avoid deeply nested IR calls for generated Set and Map multibindings causing `StackoverflowErrors` in large graphs.
```
ZacSweers:
```suggestion
- **Fix**: Avoid deeply nested IR calls for generated Set and Map multibindings causing compiler `StackOverflowError`s in large graphs at compile-time.
```

ZacSweers (approved):

PR #1588, opened 2025-12-23
Configure jdk auto installing using foojay

japplin:
I didn't have jdk 24 setup locally and claude suggested using foojay to automatically configure it's installation.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -183,7 +183,10 @@ subprojects {
 
   pluginManager.withPlugin("java") {
     configure<JavaPluginExtension> {
-      toolchain { languageVersion.set(libs.versions.jdk.map(JavaLanguageVersion::of)) }
+      toolchain {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -183,7 +183,10 @@ subprojects {
 
   pluginManager.withPlugin("java") {
     configure<JavaPluginExtension> {
-      toolchain { languageVersion.set(libs.versions.jdk.map(JavaLanguageVersion::of)) }
+      toolchain {
+        languageVersion.set(libs.versions.jdk.map(JavaLanguageVersion::of))
+        vendor.set(JvmVendorSpec.ADOPTIUM)
```
ZacSweers, line 188: will this require everyone to use adoptium? I use zulu for instance

ZacSweers:
Closing this out as it seems this would require that vendor and that's not a goal. I do have the JDK mentioned in the CONTRIBUTING docs fwiw

PR #1589, opened 2025-12-23
Fix bug preventing scope argument from being used in position other than first

japplin:
Before the fix the added testContributesBindingScopeArgumentOrder would result in a MissingBinding error. I'm not entirely sure what the original intent for these checks were but since all tests are passing without them I'm hoping they simply became redundant after other refactoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changelog
 ### Fixes
 
 - [FIR, Anvil Interop] Fix rank-based binding replacements getting dropped for multi-contribution classes in root graphs.
+- [FIR] Named annotation arguments in different order from declared parameters getting silently skipped.
 
 0.9.2
 -----


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
```diff
@@ -1147,16 +1147,6 @@ internal fun <T : Any> FirAnnotation.argumentAsOrNull(klass: Class<T>, name: Nam
 
   val argByIndex = arguments.getOrNull(index)
 
-  // Check if the index is present but under a different name
-  if (argByIndex != null && argumentMapping.mapping.isNotEmpty()) {
```
ZacSweers, line 1151: This code here was intended to be defensive cover to ensure we don't accidentally match by index but have a mismatched name from an interop'd function 🤔. I guess if this works then that works but it was supposed to match by the index in the original annotation's declaration

PR #1604, opened 2025-12-30
Ignore unused duplicate bindings

vRallev:
Delay error for duplicate bindings until they're actually used and ignore duplicate bindings when they're not used. This is similar to the behavior in Dagger.


gradle-plugin/src/functionalTest/kotlin/dev/zacsweers/metro/gradle/incremental/BindingContainerICTests.kt
```diff
@@ -1940,10 +1940,10 @@ class BindingContainerICTests : BaseIncrementalCompilationTest() {
       """
       [Metro/DuplicateBinding] Multiple bindings found for kotlin.String
 
-        test/BindingContainerA.provideString
+        test.BindingContainerA
           fun provideString(): kotlin.String
                                ~~~~~~~~~~~~~
-        test/BindingContainerB.provideString
+        test.BindingContainerB
```
ZacSweers, line 1946: gonna look at this before merging, don't wanna reduce the clarity of an error message

ZacSweers: oh actually I suppose this is less redundant, I misread before and thought it was removing the .provideString()

PR #1617, opened 2026-01-04
Integrate androidx collections to compiler artifact

ZacSweers:


ZacSweers:
After benchmarking, this change isn't worthwhile. No meaningful speed change, minimal GC change

PR #1618, opened 2026-01-04
Enable new build tools API

ZacSweers:


ZacSweers:
Will be the default in 2.3.20

PR #1635, opened 2026-01-09
Add Kotlin 2.3.20-Beta1

ZacSweers:


ZacSweers:
Needs #1621 merged first, and that PR appears to have exposed a regression in IC in kotlinc in one of the dev builds leading up to 2.3.20-Beta1. Needs more debugging to root cause it but metro's IC tests caught it.

PR #1637, opened 2026-01-09
Link IR hint files with expect/actual tracker to avoid stale hints

### LINKED ISSUE #1393: [IC] Scope change in another module is not detected
### Description

I think it is the same issue with different outcomes in different case. I am providing simplified reproductions I didn't tested. But real cases are constantly failing

### Self-contained Reproducer

**Main module**

This module depends on other modules

```kotlin
interface AppGraph

@GraphExtension(ActivityScope::class)
interface ActivityGraph {
  @GraphExtension.Factory
  fun interface Factory {
    fun create(): ActivityGraph
  }
}

@Inject
class Provider(val map: Map<KClass<*>, Any>)
```

**Case A in another module**

Change scope:
```kotlin

@Inject class Foo

@BindingContainer
@ContributesTo(ActivityScope::class)  <-- Change scope to AppScope
public interface FooModule {
  @Binds
  @IntoMap
  @ViewModelKey(Foo::class)
  public fun bindsCreditCardAddViewModel(foo: Foo): Any
}
```

In  this case duplication exception happens:
```
java.lang.IllegalStateException: Duplicate multibinding source: @dev.zacsweers.metro.internal.MultibindingElement("kotlin.reflect.KClass<out com.vinted.core.screen.BaseFragment>_com.vinted.core.screen.BaseFragment", "1937648603") com.vinted.core.screen.BaseFragment. This is a bug in the Metro compiler, please report it to https://github.com/zacsweers/metro.
	at dev.zacsweers.metro.compiler.UtilKt.reportCompilerBug(util.kt:209)
	at dev.zacsweers.metro.compiler.ir.graph.IrBinding$Multibinding.addSourceBinding(IrBinding.kt:610)
	at dev.zacsweers.metro.compiler.ir.graph.BindingGraphGenerator.generate(BindingGraphGenerator.kt:363)
```

**Case B**

Remove scope:
```kotlin

class RandomDependencyWithoutProvision

@ContributesTo(ActivityScope) <--- remove this line
@Inject class Foo(
  val dep: RandomDependencyWithoutProvision
)
```

In this case, ActivityGraph still complains about not finding how to provide `RandomDependencyWithoutProvision`

### Metro version

0.7.6

### Context

Every time it works by deleting the `build` folder on the changed module and compiling without the build cache. 

Kotlin: 2.2.21

### LINKED ISSUE #1642: Enable FIR hint gen by default
(in Kotlin 2.3.20+)

neworld:
It is test repro for #1393. In our project, we are relying on aggregation a lot, and time to time we need to adjust scopes. This also pollutes build cache, so after cleaning the build, the next complication must be without build cache. It is not only annoying, but sometimes the change to wrong scope does not report error. As for now, we decided to disable Kotlin IC to be safe. 

Anvil prevented us to enable IC due very similar issue. 

BTW, I am using `assertFail`, because build is throwing exception, and failure is expected here. 

Edit: Resolves #1393, resolves #1642

### DIFF
--- a/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/KotlinVersions.kt
+++ b/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/KotlinVersions.kt
@@ -3,9 +3,19 @@
 package dev.zacsweers.metro.gradle
 
 import org.jetbrains.kotlin.tooling.core.KotlinToolingVersion
+import org.jetbrains.kotlin.tooling.core.isDev
 
 internal object KotlinVersions {
   val kotlin230 = KotlinToolingVersion(2, 3, 0, null)
   val kotlin2320 = KotlinToolingVersion(2, 3, 20, null)
   val kotlin2320Beta1 = KotlinToolingVersion("2.3.20-Beta1")
+  val kotlin2320Dev6204 = KotlinToolingVersion("2.3.20-dev-6204")
+
+  fun supportsTopLevelFirGen(version: KotlinToolingVersion): Boolean {
+    return if (version.isDev) {
+      version >= kotlin2320Dev6204
+    } else {
+      version >= kotlin2320Beta1
+    }
+  }
 }
--- a/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroGradleSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroGradleSubplugin.kt
@@ -44,14 +44,11 @@ public class MetroGradleSubplugin @Inject constructor(problems: Problems) :
   override fun apply(target: Project) {
     val toolingVersion = target.kotlinToolingVersion
 
-    val baseToolingVersion =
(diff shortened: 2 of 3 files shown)


ZacSweers:
I took a quick glance earlier and it seems that the lib project in the first test is leaving a stale generated hint file, and thus the root project compiles. I'm not sure offhand what to do yet but will keep looking.

One thing you could do to help in your project if possible test using FIR hint generation ([details](https://github.com/ZacSweers/metro/issues/460#issuecomment-3594751377)). My hope is that metro can eventually move fully to that and I would assume it has better stale file handling out of the box.

Note you'll need to check with a dev build near the one I tested with there, as there is a breaking change in 2.3.20-Beta1 I haven't had a chance to address on main yet.

neworld:
Somehow I missed that post. Looks interesting, I will definitely try both options on our big project in a couple of days and let you know.

On a positive note, we fully migrated to Metro, and thanks to it, we now have K2 and no-kapt, which is a big win :)

ZacSweers:
Got it working!

ZacSweers:
Well, it was weird but good news is

- I have a workaround that supports <`2.3.20-Beta1`
- Switching to FIR hint gen in `2.3.20-Beta1` + fixes it forward

PR #1652, opened 2026-01-12
Switch to persistent artifacts

ZacSweers:


ZacSweers:
Will revisit after 2.3.20-Beta2

PR #1653, opened 2026-01-13
Do not report similar bindings when a missing binding has type `kotlin.Any`

JoelWilcox:
- In practice, this reported all available bindings and the list could be significant with larger graphs.


compiler-tests/src/test/data/diagnostic/aggregation/SimilarBindingsNotReportedWhenMissingBindingTypeIsAny.kt
```diff
@@ -0,0 +1,12 @@
+// RUN_PIPELINE_TILL: FIR2IR
+// RENDER_IR_DIAGNOSTICS_FULL_TEXT
+
+// MODULE: main
```
ZacSweers:
```suggestion
```

Not necessary

PR #1663, opened 2026-01-14
Add repro for Multibinding in parent class causing MethodTooLarge #1658

### LINKED ISSUE #1658: `getMemberInjector` generates too large method after 0.9.4
0.9.2 version dump was compact:
```kotlin
override val memberInjector: MembersInjector<MDActivity>
    override get(): MembersInjector<MDActivity> {
      return Companion.create(
        androidViewInjector = Companion.create(
          injectors = <this>.<get-mapOfKClass___ToInstanceInjector_View_>()
        ),
        androidFragmentInjector = Companion.create(
          injectors = <this>.<get-mapOfKClass___ToInstanceInjector_Fragment_>()
        ),
       <...>
         
```

But 0.9.4 generates full map builders:
```kotlin
  override val memberInjector: MembersInjector<MDActivity>
    override get(): MembersInjector<MDActivity> {
      return Companion.create(androidViewInjector = Companion.create(injectors = { // BLOCK
        val tmp0_builder: Builder<KClass<*>, InstanceInjector<View>> = Companion.builder<KClass<*>, InstanceInjector<View>>(size = 39)
        <.. All Views ..>
        tmp0_builder.put(key = FooView::class, providerOfValue = Companion.create(
          injector = Companion.invoke<MembersInjector<FooView>>(
            value = Companion.create(viewModelFactory = Companion.create(viewModels = { // BLOCK
              val tmp1_builder: Builder<KClass<out ViewModel>, ViewModel> = Companion.builder<KClass<out ViewModel>, ViewModel>(size = 98)
              tmp1_builder.put(key = AnotherViewModel::class, providerOfValue = Companion.create(<...>
              <.. all ViewModels ..>
```

This `<.. all ViewModels ..>` was repeated for all Views, which injected `viewModelFactory`.

Also, I found it builds map generations for many multibinded contributions, like:
```
            tmp36_builder.addCollectionProvider(collectionProvider = ProvideItemBoxFeatureExperimentMetroFactory.create())
            tmp36_builder.addCollectionProvider(collectionProvider = ProvideWebViewFeatureExperimentMetroFactory.create())
```

I am also thinking what are good way to make repro? I don't see a good place to put minimal change in `benchmark` to repro this. I think `gradle-plugin/functionalTest` will be the much easier to write and maintain, but I am not sure about performance implications and does it first functional test definition? Also, new `sample` could work, as I could port mine [metro-repro](https://github.com/neworld/metro-repro) with this case, but also with some fixed cases to have extra regressions. In all cases codegen is needed

neworld:
Original issue: #1658. It was not trivial to repro. Whatever I did, generated code was clean, until tried members injection in parent class. I think it is a culprit. I cleaned a test to only make obvious duplication. I ran `generateTest` on 0.9.2 to easier spot this duplication 

```
class MyActivity : BaseActivity()

@HasMemberInjections
abstract class BaseActivity : Activity() {
    @Inject
    lateinit var fragmentInjector: FragmentInjector
}
```

Edit: Resolves #1658

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/expressions/GraphExpressionGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/expressions/GraphExpressionGenerator.kt
@@ -7,7 +7,6 @@ import dev.zacsweers.metro.compiler.ir.IrContextualTypeKey
 import dev.zacsweers.metro.compiler.ir.IrMetroContext
 import dev.zacsweers.metro.compiler.ir.IrTypeKey
 import dev.zacsweers.metro.compiler.ir.ParentContext
-import dev.zacsweers.metro.compiler.ir.allSupertypesSequence
 import dev.zacsweers.metro.compiler.ir.graph.BindingPropertyContext
 import dev.zacsweers.metro.compiler.ir.graph.DependencyGraphNode
 import dev.zacsweers.metro.compiler.ir.graph.IrBinding
@@ -20,7 +19,6 @@ import dev.zacsweers.metro.compiler.ir.metroFunctionOf
 import dev.zacsweers.metro.compiler.ir.parameters.Parameter
 import dev.zacsweers.metro.compiler.ir.parameters.Parameters
 import dev.zacsweers.metro.compiler.ir.parameters.parameters
-import dev.zacsweers.metro.compiler.ir.rawType
 import dev.zacsweers.metro.compiler.ir.rawTypeOrNull
 import dev.zacsweers.metro.compiler.ir.regularParameters
 import dev.zacsweers.metro.compiler.ir.requireSimpleFunction


ZacSweers:
Thanks for the repro, this helped me get to the bottom of it. In short - the issue boiled down to an issue with how member injections work in multi-level. The bug here is that a MembersInjector for a subclass isn't currently depending on any membersinjectors of the parent class, and actually multibindings aren't an issue here at all but just a manifestation of that underlying issue.

I can fix this, though I'm keen to advise against relying this heavily on member injection in general as it's a bit of a code smell

neworld:
> I can fix this, though I'm keen to advise against relying this heavily on member injection in general as it's a bit of a code smell

Totally agreed. We have too much legacy code and the progress of removing MemberInjector is slower than ideally. Thankfully to Metro support, we at least have K2 and no-kapt alredy, which shows significant performance (~15%) improvements. Even without IC.

neworld:
Thanks! The dump is even better than 0.9.2

PR #1688, opened 2026-01-18
Add switching providers support

### LINKED ISSUE #42: Consider implementing SwitchingProviders

ZacSweers:
Resolves #42


ZacSweers:
TODO

- Dump tests
- Revisit if MembersInjector and CustomWrapper types can go in a switching provider

PR #1692, opened 2026-01-19
Test 2.4.0

ZacSweers:


ZacSweers:
Can't build against 2.4.x while we still base at 2.2.20. Will revisit when we raise to 2.3.0

PR #1711, opened 2026-01-22
Fix unusedGraphInputsSeverity option for graph extensions

kevinguitar:
I found out that `unusedGraphInputsSeverity` option works for DependencyGraph, but not for GraphExtension.

The problem is that the report destination points to the GraphExtensionImpl, instead of the GraphExtension declaration, so there is no way to suppress the warning.

Tried to fix it myself, but need a bit more guidance on how to locate the IR parameter at the graph extension source 🙏

### DIFF
--- a/compiler-tests/src/test/data/diagnostic/dependencygraph/UnusedGraphInputsInGraphExtensionAreReported.fir.ir.diag.txt
+++ b/compiler-tests/src/test/data/diagnostic/dependencygraph/UnusedGraphInputsInGraphExtensionAreReported.fir.ir.diag.txt
@@ -1,6 +1,9 @@
 /UnusedGraphInputsInGraphExtensionAreReported.kt:(741,759): warning: Graph input 'kotlin.Int' is unused and can be removed.
 
 
+/UnusedGraphInputsInGraphExtensionAreReported.kt:(767,803): warning: Graph input 'SomeIncludedType' is unused and can be removed.
+
+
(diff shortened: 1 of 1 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/transformers/DependencyGraphTransformer.kt
```diff
@@ -585,8 +588,19 @@ internal class DependencyGraphTransformer(
...
+              it.isAnnotatedWithAny(metroSymbols.classIds.graphExtensionAnnotations)
+            }
+          } else {
+            graphDeclaration
+          }
         reportCompat(
-          irDeclarations = sequenceOf(unusedBinding.reportableDeclaration, graphDeclaration),
+          // TODO: unusedBinding.reportableDeclaration is pointing to the graph extension impl, we
+          //  need the declaration pointing to the source.
+          irDeclarations = sequenceOf(unusedBinding.reportableDeclaration, graphDeclarationSource),
```
kevinguitar: I was able to locate the actual source of the graph extension, but `unusedBinding.reportableDeclaration` is still pointing to the graph extension impl, and I couldn't find an easy way to fix it. Looks like `IrBinding.BoundInstance.reportableDeclaration` is always pointing to the graph impl (for graph extension), and I'm not sure if this is deliberate or a potential bug for reporting 🙏

PR #1715, opened 2026-01-22
Collect exclusions from parent DependencyGraphs

### LINKED ISSUE #1713: excludes does not inherit from parent
### Description

If a DependencyGraph declares an `excludes` with items that are contributed in children GraphExtensions, they are ignored, which can causing `Metro/DuplicateBinding` errors

### Self-contained Reproducer

```kotlin
@DependencyGraph(AppScope::class, excludes = [LongBinding1::class])
interface AppGraph {
    val unitGraph: UnitGraph
}

@GraphExtension(Unit::class)
interface UnitGraph {
    val long: Long
}

@ContributesTo(Unit::class)
@BindingContainer
object LongBinding1 {
  @Provides fun provideLong(): Long = 1L
}

@ContributesTo(Unit::class)
@BindingContainer
object LongBinding2 {
  @Provides fun provideLong(): Long = 2L
}

fun box(): String {
  assertEquals(2L, graph.unitGraph.long)
  return "OK"
}
```

### Metro version

0.9.4

### Context

_No response_

jonamireh:
## Summary

This PR collects `excludes` from parent `DependencyGraph`s when computing all exclusions for a `GraphExtension`. This allows exclusions declared in the `DependencyGraph`  or `GraphExtension` to apply in any child `GraphExtension`

Fixes #1713

_Disclosure: This PR started as AI-assisted but I've changed almost all the code_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,9 @@ Changelog
 
 ### Fixes
 
+- **[IR]** Fix exclusions not inherited from parent graphs
 - **[IR/Native]** Fix parameter type equality check in native compilations in constructor-injected class factories.
 - **[IR/Compat]** Fix compatibiltiy for source-less diagnostics in Kotlin `2.3.20-Beta1`.
-- **[IR]** Fix exclusions not inherited from parent graphs
 
 0.10.0
 ------


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/IrContributionMerger.kt
```diff
@@ -62,14 +63,33 @@ internal class IrContributionMerger(
...
+        allScopes,
+        excluded + parentExcluded(callingDeclaration),
+        callingDeclaration,
+      )
     } else {
       return null
     }
   }
 
+  private fun parentExcluded(callingDeclaration: IrDeclaration): Set<ClassId> =
```
ZacSweers: this is a somewhat expensive/repeated lookup. We should def cache it at the very least, but I think we may also have this information in `GraphNode` if you have access to that in this scope?

ZacSweers: [4d85250](https://github.com/ZacSweers/metro/pull/1715/commits/4d852507f1185c7a264ecd1457e461f0d9a4cf5c)

PR #1749, opened 2026-01-27
Remove assisted factory access exception

japplin:
Square is encountering the "Assisted inject factories should only be accessed as instances" exception during compilation. This pr unsuccessfully attempts to create a failing reproducer and removes the exception.


japplin:
Resolved by https://github.com/ZacSweers/metro/pull/1756/changes

PR #1751, opened 2026-01-27
Member injection on java classes should fail compilation

C2H6O:
After migrating to Metro, I noticed that the `.java` classes that are `Component.inject(this)`'ed fail silently to inject `@Inject` annotated members. This PR adds a check that should fail during compilation.

### DIFF
--- a/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/MembersInjectErrorsTest.kt
+++ b/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/MembersInjectErrorsTest.kt
@@ -11,6 +11,67 @@ import org.junit.Test
 class MembersInjectErrorsTest : MetroCompilerTest() {
 
   @Test
+  fun `java classes cannot be used for member injection via graph inject function`() {
+    compile(
+      sourceJava(
+        """
+        public class JavaActivity {
+          @Inject String value;
+        }
+        """
+          .trimIndent()
+      ),
+      source(
+        """
+        @DependencyGraph
+        interface ExampleGraph {
+          fun inject(activity: JavaActivity)
+        }
+        """
+          .trimIndent()
+      ),
+      expectedExitCode = ExitCode.COMPILATION_ERROR,
+    ) {
+      assertDiagnostics(
+        "e: ExampleGraph.kt:8:14 Java class 'test.JavaActivity' cannot be used for member injection. Metro cannot generate injection code for Java classes. Convert the class to Kotlin or use constructor injection instead."
+      )
+    }
+  }
+
+  @Test
+  fun `java classes with injected setter cannot be used for member injection`() {
+    compile(
+      sourceJava(
+        """
+        public class JavaActivity {
+          private String value;
+
+          @Inject
+          public void setValue(String value) {
+            this.value = value;
+          }
+        }
+        """
+          .trimIndent()
+      ),
+      source(
+        """
+        @DependencyGraph
+        interface ExampleGraph {
+          fun inject(activity: JavaActivity)
+        }
+        """
+          .trimIndent()
+      ),
+      expectedExitCode = ExitCode.COMPILATION_ERROR,
+    ) {
+      assertDiagnostics(
+        "e: ExampleGraph.kt:8:14 Java class 'test.JavaActivity' cannot be used for member injection. Metro cannot generate injection code for Java classes. Convert the class to Kotlin or use constructor injection instead."
+      )
+    }
+  }
+
+  @Test
   fun `suspend functions are unsupported`() {
     compile(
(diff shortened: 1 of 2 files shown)


compiler/src/test/kotlin/dev/zacsweers/metro/compiler/fir/MembersInjectErrorsTest.kt
```diff
@@ -10,6 +10,67 @@ import org.junit.Test
 
 class MembersInjectErrorsTest : MetroCompilerTest() {
 
+  @Test
+  fun `java classes cannot be used for member injection via graph inject function`() {
```
ZacSweers, line 14: Per the CONTRIBUTING.md doc, please do not add new tests to compiler/src/test and instead use `compiler-test/`

C2H6O:
🤔 
This is how the `metro` branch is defined in our Plugin that switches between Dagger/Anvil and Metro (it does have interop):

```
"metro" -> {
                pluginManager.apply("dev.zacsweers.metro")

                afterEvaluate {
                    extensions.configure<MetroPluginExtension> {
                        interop {
                            includeDagger(includeJavax = true)
                            includeAnvilForDagger(includeJavax = true)
                        }
                    }
                }

                libs.findLibrary("google-dagger").ifPresentOrElse(
                    { dependency ->
                        dependencies.add("implementation", dependency)
                    },
                    { logger.error("Libs Error: 'dagger' not found in version catalog.") }
                )

                libs.findLibrary("zacsweers-anvil-annotations").ifPresentOrElse(
                    { dependency ->
                        dependencies.add("implementation", dependency)
                    },
                    { logger.error("Libs Error: 'zacsweers-anvil-annotations' not found in version catalog.") }
                )

                libs.findLibrary("zacsweers-metro-interop-dagger").ifPresentOrElse(
                    { dependency ->
                        dependencies.add("implementation", dependency)
                    },
                    { logger.error("Libs Error: 'zacsweers-metro-interop-dagger' not found in version catalog.") }
                )
            }
```
and the java class that gets injected by a Component still has all members as `null`

ZacSweers:
Sorry that's not really enough info to help. There are tests in metro that cover member injection to Java+dagger classes, if you have a bug then you'll need to create a minimally reproducible sample project or failing unit test

C2H6O:
Sorry, I only meant to provide that bit of information to show that we do have dagger interop enabled and the members of Android Activities that are still in .java land aren't injected when Metro is enabled. I'll try to create a failing unit test

C2H6O:
@ZacSweers here's the repro project: https://github.com/C2H6O/MetroMemberInjectionPlayground

ZacSweers:
Can you add more context? In a quick glance on mobile it doesn't look like that uses metro at all?

C2H6O:
Metro is enabled with dagger interop. There are 2 identical classes (1 Kotlin, 1 Java):

Java:
```
public class MemberInjectionJavaFragment {
    @Inject
    Resources resources;
    public void ensureResourcesNotNull() {
        System.out.println("LX+++ resources asset manager: " + resources.getAssets());
    }
}
```

Kotlin:
```
class MemberInjectionKotlinFragment {
    @Inject
    lateinit var resources: Resources
    fun ensureResourcesNonNull() {
        System.out.println("LX+++ resources asset manager: " + resources.getAssets())
    }
}
```

They both get injected through Component's `fun inject(...)` (member injection), but only Java one fails to get injected.

C2H6O:
Also, I pushed a much more simplified version of that Project. I forgot to commit/push my simplifications initially.

ZacSweers:
Ah, Dagger is not actually generating any members injectors for the java class though. You completely disabled dagger code gen, whereas Metro's interop here is that it can _use_ dagger-generated membersinjector class impls, which this isn't going to have.

ZacSweers:
Sorry wrong button. I'm open to making this PR purely about a diagnostic if interop is disabled, but the general premise here seems to be a misunderstanding about how dagger runtime interop works

C2H6O:
@ZacSweers do you mean I should keep dagger-compiler ksp dependency so that those memberinjectors get dagger-generated? I guess I got confused because step 1 of adoption (https://zacsweers.github.io/metro/latest/adoption/#option-2-migrate-existing-usages-reuse-your-existing-annotations) is to remove dagger-compiler

ZacSweers:
That option 2 doc is specific to annotation reuse only and not runtime interop (it sort of predates it). Member injection _runtime_ interop implicitly requires the dagger compiler to actually run and generate members injectors for those java classes. That's primarily for situations where you are using code compiled in an upstream module

C2H6O:
@ZacSweers okay, I think I'm going to close this PR since I don't fully grok all the pieces

PR #1753, opened 2026-01-27
Small optimizations

ZacSweers:
- Eliminate one graph pass by populating reverse adjacency during the validation
- Defer general binding validation until after population, so we only validate reachable nodes
- Use a few more primitive collection types from androidx.collection for memory reasons

### DIFF
--- a/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/graph/GraphPartitionerTest.kt
+++ b/compiler/src/test/kotlin/dev/zacsweers/metro/compiler/graph/GraphPartitionerTest.kt
@@ -267,7 +267,7 @@ class GraphPartitionerTest {
       adjacency = GraphAdjacency(adjacency, emptyMap()),
       components = components,
       componentOf =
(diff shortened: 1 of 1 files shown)


compiler/src/test/kotlin/dev/zacsweers/metro/compiler/graph/GraphPartitionerTest.kt
```diff
@@ -261,9 +264,14 @@ class GraphPartitionerTest {
       sortedKeys = sortedKeys,
       deferredTypes = emptySet(),
       reachableKeys = adjacency.keys.toSet(),
-      adjacency = adjacency,
+      adjacency = GraphAdjacency(adjacency, emptyMap()),
       components = components,
-      componentOf = componentOf,
-      componentDag = emptyMap(),
+      componentOf =
+        MutableObjectIntMap<String>().apply {
```
ZacSweers:
```suggestion
        MutableObjectIntMap<String>(componentOf.size).apply {
```

PR #1754, opened 2026-01-28
Fix inherited bindings duplicate binding error

C2H6O:
Fixes the bug discussed offline


ZacSweers:
I don't think this is the right fix tbh, it's a bit of a bandaid over something that seems to be the result of some logic farther up the stack. I'm going to look further into this separately

C2H6O:
Feel free to close this PR

ZacSweers:
Figured this out!

PR #1771, opened 2026-01-31
Add regression test for multi-module stale IC contribution with binding containers

heorhiipopov:
A reproduction test case demonstrating an IC bug reported in https://github.com/ZacSweers/metro/issues/1772


ZacSweers:
Merging this finally, only 2.3.20+

PR #1785, opened 2026-02-07
Fix/cycle reporter accuracy

heorhiipopov:
The cycle reporter used the raw list of nodes in an SCC to generate the diagnostic trace. An SCC can contain nodes that are _not_ connected by "hard" (scalar) dependencies.

When the reporter attempted to construct a trace, it assumed a path through the SCC nodes. If it encountered a link where no direct scalar dependency existed, it would fail to find the corresponding value parameter and throw an `IllegalStateException: No value parameter found...`. Noticed this when a circular dependency with a deep and diverse trace was introduced in my project. 

Here, trying to fix this crash by using BFS instead of relying on SCC node list. New reporter will follow "hard" SCC edges(ones that are not `isDeferrable` and not `isImplicitlyDeferrable`), find minimal "hard" path that closes the loop, and reconstruct the trace in a `Requestor` -> `Dependency` order, and it should never hit something that does not exist. 

Added a test too to verify that it works correctly. I can submit a functional test that reproduces the issue separately if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Special thanks to the following contributors for contributing to this release!
 
 - [@kevinguitar](https://github.com/kevinguitar)
 - [@DaniilPavlenko](https://github.com/DaniilPavlenko)
+- [@heorhiipopov](https://github.com/heorhiipopov)
 
 0.10.2
 ------


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/graph/BindingGraph.kt
```diff
@@ -278,31 +278,39 @@ internal open class MutableBindingGraph<
...
+
+            val finalPath = cyclePath ?: sccVertices
+            val entriesInCycle = buildList {
+              val size = finalPath.size
+              for (i in 0..size) {
+                val currentDep = finalPath[i % size]
+                val prevReq = if (i == 0) finalPath.last() else finalPath[i - 1]
+                val callingBinding = bindings.getValue(prevReq)
+                val contextKey =
+                  callingBinding.dependencies.first { it.typeKey == currentDep && !it.isDeferrable }
```
ZacSweers: This'll still end up hitting the NSEE potentially right? Or if we know in this case that it should never happen, let's do something like `firstOrNull {} ?: reportCompilerBug("something more specific about why this should never happen")`

heorhiipopov: Could happen if there is something wrong with `isHardEdge`, though now I can't think of any scenario where this check would not be enough. 
Will report a compiler bug in this case 👍

heorhiipopov:
>A functional test would be great too, there's a reports test category in compiler-tests that would be good for this and can verify contents in reports as a golden image.

Sure, I can add some later, also seen a couple of tests fail for 2.3.20 betas, will take a look into that too

heorhiipopov:
Interesting thing about tests(at least for me 😆 ): on 2.3.0, no matter what I put in expected output in `testMixedCycleParams`, test still passes. It only fails if I remove `<!METRO_ERROR!>...<!>` tag. 

But on 2.3.20 betas, tests actually seem to care about expected output. So I had to adjust the expected output cycle of `testMixedCycleParams` from:
```
RealA
    --> A
    ~~> FakeA
    --> B
    --> RealA    // 4 nodes
```

To:
`B --> A ~~> FakeA --> B //3 nodes`

We report a different path now because BFS conveniently finds the shortest one.

As far as I understand, in `MixedCyclesParams` we can break both cycles: longer one and shorter one, by making `FakeA`'s dependency on `B` `Lazy`. In this case I do not see a way to fix only one cycle without fixing another one. So I think the changes I made to that test make sense, given the way new reporter works. 

As for the generated diagnostic test for repro/regression purposes, it's kinda weird too. Looks like with Kotlin 2.3.0 we don't hit whatever we need to hit in these tests and the crash does not manifest when I try to reproduce. Moreover, looks like even a cycle is not found there, because when I compile the following snippet in these tests:
```
interface A
interface B
interface C

@BindingContainer
object MyModule {
    // If the reporter tries to link A -> C directly, it will crash.

    @Provides fun provideA(b: B): A = object : A {}
    @Provides fun provideB(a: A, c: Lazy<C>): B = object : B {}
    @Provides fun provideC(b: B): C = object : C {}
}

@DependencyGraph
interface AppGraph {
    val a: A
    @DependencyGraph.Factory
    interface Factory {
        fun create(@Includes module: MyModule): AppGraph
    }
}
```

It passes without error, no matter what `onCycle` logic is in `MutableBindingGraph` - one that I introduce in this PR, or the one that's on main right now. 

But on `2.3.20-Beta1` or `Beta2` we hit whatever we need to hit and it's reproducible. On these versions in shows both the `IllegalStateException` without improved `onCycle` and reporting, and hits the cycle with improved `onCycle` and reports it without a crash.

It's always reproducible with gradle testKit though, no matter what `metro.testCompilerVersion` I set to locally, so I've created a `DiagnosticsTest` in `:gradle-plugin` and added a repro/regression test there.

ZacSweers:
> It only fails if I remove <!METRO_ERROR!>...<!> tag.

For future ref you don't need to manually edit these if you use the compiler dev kit plugin that the README mentions!

And yes in 2.3.20 there appears to be a bugfix for stale tests. Previous tests do care about outputs, but in some scenarios it didn't always fail if it changed if there wasn't some other trigger it seems.

It's quite surprising that there's _no_ failure on 2.3.0 since this isn't really logic specific to the kotlin compiler right? I'm fine with starting with the gradle-based integration test but def don't want it to stay there long term as that'll likely only ever cover JVM and is really just for IC tests.

heorhiipopov:
>this isn't really logic specific to the kotlin compiler right

Yep, I think this logic should work the same on all compiler versions(as current `testKit` test shows)

I don't see a way to move this test right now, as `2.3.0` just does not fail, unless it will be possible to run this test only for `2.3.20-Beta1` and forward. But I'm also not very ok with selectively running tests for certain versions when I expect the behaviour to be similar across all versions 🤷 

I'll try to look deeper into this

ZacSweers:
It repro'd fine for me on 2.3.0 so I went ahead and moved it back

heorhiipopov:
>It repro'd fine for me on 2.3.0 so I went ahead and moved it back

Thanks for taking care of it!

PR #1808, opened 2026-02-11
Drop FIR-in-IR work-around for repeated annotations

JoelWilcox:
- Original issue: https://youtrack.jetbrains.com/issue/KT-83185
- Might be a little early since the main build is on 2.3.0 still, just drafting for now


JoelWilcox:
2.3.20 and 2.4 look good. Closing since it'll probably still be a bit before this makes sense

Issue #1816, opened 2026-02-13
ClassCastException on runtime when providing a Map


inorichi:
### Description

The attached reproducer compiles fine but produces the following crash on runtime: 
`java.lang.ClassCastException: foo.FooComponent$Impl cannot be cast to java.util.Map`

This happens on Metro >= 0.10.0.
On Metro 0.9.4 there's a codegen exception, but below 0.9.4 it works as it should.

### Self-contained Reproducer

```kt
package foo

import android.app.Application
import dev.zacsweers.metro.BindingContainer
import dev.zacsweers.metro.Binds
import dev.zacsweers.metro.DependencyGraph
import dev.zacsweers.metro.Inject
import dev.zacsweers.metro.Provides
import dev.zacsweers.metro.createGraphFactory

class FooApp : Application() {
    override fun onCreate() {
        super.onCreate()
        createGraphFactory<FooComponent.Factory>()
            .create(bar = emptyMap())
            .foo() // ClassCastException
    }
}

@DependencyGraph(bindingContainers = [FooModule::class])
interface FooComponent {
    fun foo(): Foo

    @DependencyGraph.Factory
    interface Factory {
        fun create(@Provides bar: Map<String, String>): FooComponent
    }
}

@BindingContainer
interface FooModule {
    @Binds
    fun bindFoo(impl: FooImpl): Foo
}

interface Foo

@Inject
class FooImpl(bar: Map<String, String>) : Foo
```

### Metro version

0.10.0-0.10.3

### Previous working version

0.9.3

### Context

Runtime bug on Android with Kotlin 2.3.10

inorichi:
I just noticed this is very similar to #1690 which I reported. I think that time I just compiled it but never tried to actually execute the app (I still haven't fully switched to metro).

ZacSweers:
Can you make a failing unit test in compiler-tests box tests? Given I wasn't able to reproduce your issue last time. The README there should be pretty straightforward, and would like to get the Android classes out of your repro. If you open a PR with that failing test it gives me an easy foothold to work against

ZacSweers:
Also, please include the full stacktrace

inorichi:
Same thing happens if I turn this module into a JVM application with the following main:

```kt
fun main() {
    createGraphFactory<FooComponent.Factory>()
        .create(bar = emptyMap())
        .foo() // ClassCastException
}
```

The full exception is:

```
[...]

PR #1823, opened 2026-02-14
Fix token gen for direct map inputs to extensions

inorichi:
This PR adds a test case for a new compiler error on Metro 0.10.4 after #1816 was fixed.

### DIFF
--- a/compiler-tests/src/test/data/box/contributesgraphextension/ProvidesMapIntoGraphExtensionAsLateinit.kt
+++ b/compiler-tests/src/test/data/box/contributesgraphextension/ProvidesMapIntoGraphExtensionAsLateinit.kt
@@ -1,33 +1,26 @@
-// MODULE: lib
 @DependencyGraph
-interface FooComponent {
-  fun subcomponent(): FooSubcomponent
+interface AppGraph {
+  fun loggedInGraph(): LoggedInGraph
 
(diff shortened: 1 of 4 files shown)


compiler-tests/src/test/data/box/contributesgraphextension/ProvidesMapIntoGraphExtensionAsLateinit.kt
```diff
@@ -1,33 +1,26 @@
...
 
 @GraphExtension
-interface FooSubcomponent {
-  fun inject(activity: Activity)
+interface LoggedInGraph {
+  val example: Example
 }
 
-class Activity {
-  @Inject
```
inorichi, line 20: The filename `ProvidesMapIntoGraphExtensionAsLateinit.kt` loses its meaning with this change. I really thought it only happened on lateinit injections but I may be wrong.

ZacSweers: yah can fix, and no it's not unique to lateinits in my testing

ZacSweers: #1825

PR #1840, opened 2026-02-18
Add a compiler test for generating `@Provides` and `@Binds` functions in extensions

vRallev:
This adds a compiler test for Metro extensions that generate an inner interface for custom annotations with a `@Provides` and `@Binds` functions using `MetroContributionExtension` and `MetroFirDeclarationGenerationExtension`. While the API of these interfaces is not stable and may break, extensions usually have the use of generating their own containers with a `@Binds` or `@Provides` function.


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ProvidesFactoryFirGenerator.kt
```diff
@@ -250,6 +250,24 @@ internal class ProvidesFactoryFirGenerator(session: FirSession, compatContext: C
...
+          // For source-declared @Provides functions with unresolved return types
+          // (FirUserTypeRef), ProvidesFactorySupertypeGenerator still handles them as before.
+          @OptIn(SymbolInternals::class) val returnTypeRef = sourceCallable.symbol.fir.returnTypeRef
+          if (returnTypeRef is FirResolvedTypeRef) {
+            val factoryType =
+              session.symbolProvider
+                .getClassLikeSymbolByClassId(Symbols.ClassIds.metroFactory)!!
+                .constructType(arrayOf(returnTypeRef.coneType))
+            superType(factoryType)
+          }
```
vRallev, line 270: To be clear, I'm not married to any of the tests above, but they verify our use cases. That is helpful for us. 

However, this fix is critical and blocks our extensions right now.

ZacSweers: Do we need to gate supertype generation later to not run if it's already set here? Alternatively, is this something metro's composite generation should handle to make sure the appropriate callback happens?

vRallev: In this particular case it should not happen in a callback. The extension generates the `@Provides` method and Metro should handle generating the Factory for that. The extension shouldn't need to deal with this part. But this is where the failure happens. 

> Do we need to gate supertype generation later to not run if it's already set here?

I'm not sure. It doesn't seem to be needed.

ZacSweers: I guess it's a bridge we can cross if we get there. I do still wonder if this is something that should be pushed out to an extension in the future but I'll let you know if I make any changes around that

PR #1845, opened 2026-02-19
Restructure reports to use hierarchical nesting instead of top-level concatenated names

JoelWilcox:
- This fixes 'file name too long' exceptions when generating reports for deeply nested graphs. For example, the report file `reports/keys-populated-test_Graph_ChildGraph.txt` will now be generated as `reports/keys-populated/test/Graph/ChildGraph.txt`
- It also protects against other variations of this exception in the future

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirBuiltIns.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirBuiltIns.kt
@@ -5,6 +5,7 @@ package dev.zacsweers.metro.compiler.fir
 import dev.zacsweers.metro.compiler.ClassIds
 import dev.zacsweers.metro.compiler.MetroOptions
 import dev.zacsweers.metro.compiler.compat.CompatContext
+import dev.zacsweers.metro.compiler.createDiagnosticReportPath
 import dev.zacsweers.metro.compiler.memoize
 import dev.zacsweers.metro.compiler.symbols.Symbols
 import java.nio.file.Path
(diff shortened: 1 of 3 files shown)


compiler-tests/src/test/data/dump/reports/DeeplyNestedGraphsKeysPopulatedReport.kt
```diff
@@ -0,0 +1,44 @@
+// This test is complementary to the [DeeplyNestedGraphsAreHandledWhenGeneratingReports] box test,
+// which verifies a sufficiently nested graph structure does not run into 'file name too long'
+// exceptions for the report files. We are not able to test the same behavior here because defining
+// a sufficiently nested graph in a dump test results in a 'file name too long' exception for an
+// equivalent '.class' file reference in the generated Java [ReportsTestGenerated] file.
```
JoelWilcox, line 5: Initially I was thinking this test could serve both purposes (regression coverage and verifying actual report structure/content) but ran into the issue described here. Link to previous CI run showing this error: https://github.com/ZacSweers/metro/actions/runs/22163119878/job/64084317480?pr=1845.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/MetroFirBuiltIns.kt
```diff
@@ -162,15 +162,29 @@ internal class MetroFirBuiltIns(
...
-  private fun writeDiagnostic(reportsDir: Path, fileName: String, text: String) {
+  private fun writeDiagnostic(
+    reportsDir: Path,
+    diagnosticKey: String,
+    fileName: String,
+    text: String,
+  ) {
     reportsDir
-      .resolve(fileName)
+      .resolve("$diagnosticKey/${fileName.replace(Regex("[-_]"), "/")}")
```
ZacSweers: nit: extract this compiled regex somewhere static so it's not recompiled every time

JoelWilcox:
> Can you add a note to the CHANGELOG?

Yes! I included a line here https://github.com/ZacSweers/metro/pull/1845/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed but GitHub hid it at the end of the changelist. Let me know if you meant in addition to this

ZacSweers:
d'oh I missed that. Thanks!

PR #1850, opened 2026-02-19
Docs/add vinted metro success story

heorhiipopov:
Hi! We've finished adopting Metro a while ago and now we're excited to share our success story. I've added a link to the article summarising our migration experience, and a short summary of the most important benefits we've gained after adopting Metro. We'd be grateful if you could give it a read and perhaps add it to the docs 😃


ZacSweers (commented):
Thanks! Nice write up. I think it might be helpful to add context more explicitly in the build results around existing code gen like your KSP processors. A lot of metro's biggest successes are in cases where the codebase goes from kapt to no source gen, but middle-ground scenarios where there is still KSP will be less significant differences

PR #1851, opened 2026-02-19
Add `getContributionHints()` API to `MetroFirDeclarationGenerationExtension`

vRallev:
`ContributionHintFirGenerator` uses predicate-based lookup which only finds source declarations, missing classes generated by external FIR extensions. This adds a `getContributionHints()` method so extensions can declare their generated contributions for hint generation.

The idea behind this new API is similar to `MetroContributionExtension`.

### DIFF
--- a/compiler/API.md
+++ b/compiler/API.md
@@ -72,6 +72,27 @@ Use this extension to generate FIR declarations (classes, functions, properties)
 
 This API is similar to `FirDeclarationGenerationExtension`, but has a `ServiceLoader` API to be loaded by Metro at compile-time and handles all its callbacks.
 
+#### `getContributionHints()`
+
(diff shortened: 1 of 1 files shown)


compiler-tests/src/test/data/box/reports/DeeplyNestedGraphsAreHandledWhenGeneratingReports.kt
```diff
@@ -1,55 +0,0 @@
-// Similar to https://github.com/ZacSweers/metro/issues/1303
-/**
```
vRallev: I had to delete this test temporarily until Joel fixes the issue. Ignore this commit.

vRallev:
> Can you add a bit to the API.md doc for posterity?

Done

PR #1852, opened 2026-02-19
Use system path separator when creating diagnostic report paths

JoelWilcox:
In particular, this addresses issues on Windows compared to always using forward slashes


JoelWilcox:
Original PR needed to be reverted, will be rolling related changes from here into the re-do

PR #1855, opened 2026-02-20
Fix Kotlin 2.3.20 compatibility for external contribution discovery

vRallev:
Two issues when compiling against Kotlin 2.3.20-RC:

1. `ExternalMetroFirTypeResolver` crashes with "Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl" during SUPERTYPES phase. In 2.3.20, coneTypeIfResolved() returns null for external compiled classes, causing `resolveClassId()` to fall through to `resolvedArgumentTypeRef()` which constructs a fresh unresolved `FirUserTypeRef`. Changed `ExternalMetroFirTypeResolver` from a stateless singleton to a session-aware class that tries `coneTypeOrNull` first and falls back to `session.typeResolver` for unresolved refs.

2. "A supertype appears twice" when the same `MetroContribution` type is discovered via both hint-based and `MetroContributionExtension` paths under different TreeMap keys. Added `.distinctBy { it.classId }` to deduplicate the final supertype list.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
@@ -57,12 +57,6 @@ internal sealed interface MetroFirTypeResolver {
 
   private class ExternalMetroFirTypeResolver(private val session: FirSession) :
     MetroFirTypeResolver {
-    private val configuration =
-      TypeResolutionConfiguration(
-        scopes = emptyList(),
-        containingClassDeclarations = emptyList(),
-        useSiteFile = null,
-      )
 
     override fun resolveType(typeRef: FirTypeRef): ConeKotlinType {
       check(typeRef is FirUserTypeRef)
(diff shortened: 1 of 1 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
```diff
@@ -53,10 +55,34 @@ internal sealed interface MetroFirTypeResolver {
...
   }
 
-  private object ExternalMetroFirTypeResolver : MetroFirTypeResolver {
+  private class ExternalMetroFirTypeResolver(private val session: FirSession) :
+    MetroFirTypeResolver {
+    private val configuration =
+      TypeResolutionConfiguration(
+        scopes = emptyList(),
+        containingClassDeclarations = emptyList(),
+        useSiteFile = null,
```
ZacSweers: Is this just a throwaway instance? If so could we keep a singleton somewhere?

ZacSweers (approved):
One nit before I merge!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/TypeResolverFactory.kt
vRallev: Done

PR #1863, opened 2026-02-20
Adds tests for and fixes two issues found in squares android monorepo

japplin:
These changes are pure claude vibes. My workflow was tell claude to create reproducer project -> then add failing test based on the reproducer project -> fix failing test in metro. After repeating that loop twice our large point of sale apps  started building.


ZacSweers:
oop didn't realize it was targeting the other branch. Will cherrypick

PR #1865, opened 2026-02-20
Restructure reports to use hierarchical nesting instead of top-level concatenated names

JoelWilcox:
- Round 2 for this one to resolve test issues
- This fixes 'file name too long' exceptions when generating reports for deeply nested graphs. For example, the report file `reports/keys-populated-test_Graph_ChildGraph.txt` will now be generated as `reports/keys-populated/test/Graph/ChildGraph.txt`
- It also protects against other variations of this exception in the future


JoelWilcox:
Passing build with all multiplatform shards included: https://github.com/ZacSweers/metro/actions/runs/22240714781/job/64343393051?pr=1865

PR #1871, opened 2026-02-22
Introduce `@ContributesTemplate` meta-annotation

japplin:
`@ContributesTemplate` is a new meta-annotation that lets users define custom annotations whose companion `@Provides` functions serve as templates. For each class annotated with the custom annotation, Metro generates a nested binding container object that stamps out the companion's `@Provides` functions with the target class substituted for the type parameter `T`.

## How it works

1. Define a custom annotation meta-annotated with `@ContributesTemplate`. Its companion object holds template `@Provides` functions, each with exactly one type parameter `T` representing the eventual target type.
2. Annotate target classes (`@Inject`, `object`, or `interface`) with the custom annotation.
3. At compile time Metro generates a `@BindingContainer` object per target that delegates back to the companion's functions with the concrete type.

## Key features

- **`TemplateScope`**: A placeholder scope marker usable in `@SingleIn(TemplateScope::class)` inside companion `@Provides` functions. Metro substitutes it at compile time with the resolved scope.
- **Upper-bound validation**: Type parameter bounds on companion `@Provides` functions (e.g. `<T : Feature>`) are validated at compile time against each target class.
- **`replaces` support**: Generated containers participate in the standard `replaces` mechanism, targeting the annotated class itself.
- **`inline reified` support**: Template functions can be `inline reified` to access the target class at runtime.
- **Cross-module support**: Custom annotations and their targets can live in different compilation units.
- **Repeated annotations**: A class can be annotated with multiple `@ContributesTemplate`-based annotations (or the same annotation multiple times with different scopes).

## Compile-time validation (`ContributesTemplateChecker`)

Validates both the annotation definition and each usage site:

- Annotation must have a companion object with at least one `@Provides` function.
- Each companion `@Provides` function must have exactly one type parameter.
- Annotation must have a `scope: KClass<*>` parameter or the meta-annotation must specify a `scope`.
- Target classes must satisfy all upper bounds declared on `T`.


runtime/src/commonMain/kotlin/dev/zacsweers/metro/ContributesTemplate.kt
```diff
@@ -0,0 +1,71 @@
...
+ * @ContributesTemplate
+ * annotation class CustomAnnotation(val scope: KClass<*>, val replaces: Array<KClass<*>> = []) {
+ *   companion object {
+ *     @Provides @IntoSet fun <T : Any> provideIntoSet(target: T): Any = target
+ *   }
+ * }
+ *
+ * // 2. Apply it to target classes
+ * @CustomAnnotation(AppScope::class)
+ * class TargetClass @Inject constructor()
```
vRallev, line 29: Should be an `@Inject` on the class.

runtime/src/commonMain/kotlin/dev/zacsweers/metro/ContributesTemplate.kt
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * ## TemplateScope
+ *
+ * Companion `@Provides` functions can use `@SingleIn(TemplateScope::class)` and the compiler will
+ * substitute it with the actual scope at compile time.
+ *
+ * ## Requirements
+ * - The annotated annotation must have a `scope: KClass<*>` parameter, or [scope] must be set on
+ *   `@ContributesTemplate`. These are mutually exclusive.
+ * - The annotation must have a companion object with at least one `@Provides` function.
```
vRallev: `@Binds` is not possible? This would be cheaper in various cases, in particular your example.

japplin: This is a shortcoming with the current design and it's unfortunate since `@Binds` is such a prominent use case for custom annotations. I was trying to keep the api as simple as possible but perhaps it's worth figure out how to support `@Binds`. WDYT of this approach:

```
@Binds fun <T : Scoped> myScoped(target: T): Scoped = error("Must be implemented by Metro.")
```

We will then have to add special casing to handle these companion object `@Binds`.

vRallev: I could live with this option, but it's a little awkward. Curious what Zac thinks.

ZacSweers: https://github.com/ZacSweers/metro/discussions/1836#discussioncomment-15890142

ZacSweers:
Closing this for now until this work can be picked up again in the future

PR #1873, opened 2026-02-22
Fix IC issue with non-assisted param removed from constructor

Egorand:
Note: Both the test case and the solution are AI-generated with small manual tweaks and cleanup. I spent some time reading the code and I believe I understand what's causing the issue, that said, I don't fully comprehend the solution as I'm not familiar with IC mechanics - I'm really sorry if the solution doesn't make sense. I verified that the integration test fails without the fix and passes with the fix, plus I cherry-picked this fix on top of the 0.10.4 tag and tested it in our app to verify that it fixes our issue. More about the issue below.

---

Our app setup is roughly as follows:

```kotlin
// :presenters Gradle module
class Presenter @AssistedInject constructor(
  private val analytics: Analytics,
  private val featureFlagManager: FeatureFlagManager,
  ...
  @Assisted private val navigator: Navigator,
) : MoleculePresenter<ViewModel, ViewEvent> {
  ...
  @AssistedFactory
  interface Factory {
    fun create(navigator: Navigator): Presenter
  }
}  

// :applets Gradle module, has an "api projects.presenters" dependency
@BindingContainer
@ContributesTo(SandboxedActivityScope::class)
class AppletViewsModule {
  companion object {
    @Provides @IntoSet
    fun balanceApplet(presenterFactory: Presenter.Factory): Applet {
      return Applet(
        ...
        tileBuilder = { navigator, scope ->
          val presenter =
            presenterFactory.create(navigator).asPresenter().start(scope, lifecycleOwner)
          ...
        },
      )
    }
  }
}

// :app Gradle module that depends on "projects.applets" and creates a graph for SandboxedActivityScope
```

We discovered that removing or adding a non-assisted parameter to `Presenter`'s constructor, e.g. as follows:

```kotlin
class Presenter @AssistedInject constructor(
  private val analytics: Analytics,
  // private val featureFlagManager: FeatureFlagManager,
  ...
  @Assisted private val navigator: Navigator,
) : MoleculePresenter<ViewModel, ViewEvent> {
  ...
  @AssistedFactory
  interface Factory {
    fun create(navigator: Navigator): Presenter
  }
} 
```

Leads to a runtime crash that looks roughly like this:

```kotlin
java.lang.NoSuchMethodError: No virtual method create(Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;Ldev/zacsweers/metro/Provider;)Lcom/squareup/cash/.../presenters/Presenter$MetroFactory; in class Lcom/squareup/cash/.../presenters/Presenter$MetroFactory$Companion; or its super classes (declaration of 'com.squareup.cash.presenters.Presenter$MetroFactory$Companion' appears in /data/app/~~IbWdTUh4tVmh7PLzPcGeqg==/com.squareup.cash.beta.debug-8r5U_uv532p-ul6LsNlTHA==/base.apk!classes13.dex)
  at com.squareup.cash.VariantSandboxedComponent$Impl$SandboxedActivityComponentImpl.getSetOfAppletProvider(Unknown Source:190)
```

There seems to be a missing IC dependency between "sourceGraph" and the "create" function that Metro generates for the `@AssistedInject`-ed class, so when the function's signature changes the graph doesn't get recompiled. The fix addresses it by tracking that function as a graph dependency.

### DIFF
--- a/gradle-plugin/src/functionalTest/kotlin/dev/zacsweers/metro/gradle/incremental/ICTests.kt
+++ b/gradle-plugin/src/functionalTest/kotlin/dev/zacsweers/metro/gradle/incremental/ICTests.kt
@@ -2737,4 +2737,149 @@ class ICTests : BaseIncrementalCompilationTest() {
     assertThat(secondBuildResult.task(":compileKotlin")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
     assertThat(project.invokeMain<String>()).isEqualTo("Hello, world42")
   }
+
+  /**
+   * Tests that removing a non-assisted parameter from an @AssistedInject class' constructor is
+   * correctly detected during incremental compilation when the factory is contributed into a set
+   * via a @BindingContainer.
+   *
+   * The three-module layout is critical:
+   * - `lib` owns AssistedClass (@AssistedInject with multiple non-assisted params).
+   * - `middle` owns BaseFactory and the @BindingContainer that @Provides @IntoSet the factory; its
+   *   ABI does NOT change when AssistedClass loses a constructor param because
+   *   AssistedClass.Factory (the interface) is unchanged.
+   * - `root` depends only on `middle` (directly), so Metro reads AssistedClass metadata during
+   *   root's recompilation from a stale cache and regenerates AppGraph$Impl with the old Provider
+   *   arity, producing a NoSuchMethodError at runtime.
+   */
+  @Test
+  fun `removing non-assisted param from an assisted inject class is detected in IC`() {
+    val fixture =
+      object : MetroProject() {
+        override fun buildGradleProject() = multiModuleProject {
+          root {
+            sources(graphAndMain)
+            dependencies(implementation(":middle"))
+          }
+          subproject("middle") {
+            sources(baseFactory, assistedModule)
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/functionalTest/kotlin/dev/zacsweers/metro/gradle/incremental/ICTests.kt
```diff
@@ -2737,4 +2737,149 @@ class ICTests : BaseIncrementalCompilationTest() {
     assertThat(secondBuildResult.task(":compileKotlin")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
     assertThat(project.invokeMain<String>()).isEqualTo("Hello, world42")
   }
+
+  /**
```
Egorand, line 2741: I'm not sure how much of these comments is useful, it did help me understand the issue better, but happy to delete some (or all) if they feel redundant.

PR #1876, opened 2026-02-23
Support external FIR extensions generating `@DependencyGraph` classes

vRallev:
Two fixes to allow `MetroFirDeclarationGenerationExtension` implementations to generate classes annotated with `@DependencyGraph`:

1. `DependencyGraphFirGenerator.findCreator`: Fall back to scope-based lookup when searching for `@DependencyGraph.Factory` nested classes. Classes generated by external FIR extensions are visible in the declared member scope but not in `declarationSymbols`, so the existing declarationSymbols-only search missed them.

2. `ContributedInterfaceSupertypeGenerator`: Two changes to support contribution merging (`@ContributesTo`, etc.) into generated graphs:

   a. needTransformSupertypes: Removed the dependencyGraphs predicate-based set gate. The predicate-based provider only contains source declarations, so generated `@DependencyGraph` classes were never matched. Now checks the annotation directly, which is functionally equivalent for source classes (the annotation check was already the next line) and additionally supports generated classes.

   b. Added `computeAdditionalSupertypesForGeneratedNestedClass` override. FIR does not call `computeAdditionalSupertypes` for classes produced by `FirDeclarationGenerationExtension`, it calls the separate computeAdditionalSupertypesForGeneratedNestedClass API instead. Without this override, `@ContributesTo` contributions were never merged into generated dependency graphs.

Includes a box test `GeneratedDependencyGraph` that exercises all three fixes end-to-end.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
@@ -37,6 +37,7 @@ import org.jetbrains.kotlin.fir.declarations.FirClass
 import org.jetbrains.kotlin.fir.declarations.FirClassLikeDeclaration
 import org.jetbrains.kotlin.fir.declarations.FirDeclarationOrigin
 import org.jetbrains.kotlin.fir.declarations.FirMemberDeclaration
+import org.jetbrains.kotlin.fir.declarations.FirResolvePhase
 import org.jetbrains.kotlin.fir.declarations.FirTypeParameterRef
 import org.jetbrains.kotlin.fir.declarations.evaluateAs
 import org.jetbrains.kotlin.fir.declarations.findArgumentByName
@@ -1326,12 +1327,18 @@ private class ConeIdRendererForDiagnostics : ConeIdRenderer() {
 }
 
 context(context: CheckerContext)
-internal fun FirClassSymbol<*>.nestedClasses(): List<FirRegularClassSymbol> =
-  nestedClasses(context.session)
+internal fun FirClassSymbol<*>.nestedClasses(
+  memberRequiredPhase: FirResolvePhase = FirResolvePhase.STATUS
+): List<FirRegularClassSymbol> =
+  nestedClasses(context.session, memberRequiredPhase = memberRequiredPhase)
 
-internal fun FirClassSymbol<*>.nestedClasses(session: FirSession): List<FirRegularClassSymbol> {
+internal fun FirClassSymbol<*>.nestedClasses(
+  session: FirSession,
+  memberRequiredPhase: FirResolvePhase = FirResolvePhase.STATUS,
+): List<FirRegularClassSymbol> {
   val collected = mutableListOf<FirRegularClassSymbol>()
-  declaredMemberScope(session, memberRequiredPhase = null).processAllClassifiers { symbol ->
+  declaredMemberScope(session, memberRequiredPhase = memberRequiredPhase).processAllClassifiers {
+    symbol ->
     if (symbol is FirRegularClassSymbol) {
       collected += symbol
     }


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -265,6 +260,27 @@ internal class ContributedInterfaceSupertypeGenerator(
...
+    // If there is no scope argument, then nothing needs to be merged.
+    if (graphAnnotation.scopeArgument() == null) {
+      return emptyList()
+    }
+
+    return computeAdditionalSupertypes(
+      klass,
+      klass.superTypeRefs.filterIsInstance<FirResolvedTypeRef>(),
+      typeResolver,
+    )
```
ZacSweers: Let's extract a common impl function for both of these so we don't double lookup the graph annotation or double the scope checks

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
```diff
@@ -556,8 +558,25 @@ internal class DependencyGraphFirGenerator(session: FirSession, compatContext: C
...
-      log("Creator found from $context? $creator.")
-      return creator
+      if (creator != null) {
+        log("Creator found from $context? $creator.")
+        return creator
+      }
+
+      // Fall back to scope-based lookup for classifiers generated by external FIR extensions
+      // (which are visible in scope but not in declarationSymbols).
+      val scope = classSymbol.declaredMemberScope(session, memberRequiredPhase = null)
```
ZacSweers: Can we use the `nestedClasses()` util function for this in fir.kt? Or does that run to too distant of a phase to be usable here?

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
vRallev: Ah, I missed that. Yes, I made the change.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/DependencyGraphFirGenerator.kt
vRallev: Yes, that works!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
```diff
@@ -1325,9 +1326,12 @@ private class ConeIdRendererForDiagnostics : ConeIdRenderer() {
...
 
 context(context: CheckerContext)
-internal fun FirClassSymbol<*>.nestedClasses(): List<FirRegularClassSymbol> {
+internal fun FirClassSymbol<*>.nestedClasses(): List<FirRegularClassSymbol> =
+  nestedClasses(context.session)
+
+internal fun FirClassSymbol<*>.nestedClasses(session: FirSession): List<FirRegularClassSymbol> {
   val collected = mutableListOf<FirRegularClassSymbol>()
-  declaredMemberScope().processAllClassifiers { symbol ->
+  declaredMemberScope(session, memberRequiredPhase = null).processAllClassifiers { symbol ->
```
ZacSweers: nit: this is changing semantics slightly from before (which ran to `STATUS`). Any implications/concerns?

vRallev: I honestly don't know and rely a lot on the test coverage here. It didn't make a difference for any of our tests. For my own sanity I asked Claude and it said it doesn't matter here, because we're only processing class symbols and don't rely resolved members or properties. But this sounds like a ticking time bomb? I can revert this, what do you think?

vRallev: I can also add parameter and set it to null by default. If we don't need resolved members, then I'm not sure if we should request that phase.

ZacSweers: Let's add a parameter but let's have it match the previous behavior (STATUS) rather than null just to be safe. Where this runs into issues usually is IDE, which unfortunately doesn't have as much coverage

vRallev: Done

ZacSweers (approved):

PR #1893, opened 2026-02-25
Add sample instrumented test for android

rubensousa:
- Added simple test to the android sample that just asserts that a fake instance of a tracker received an event
- Test replaces the app graph with a newly created one with a fake binding, to allow replacing the tracker implementation.

**However**: For some reason, the ViewModel is still receiving the real `TrackerImpl`, not the `FakeTracker`. Would need support here, because I'm still getting familiar with the APIs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,7 @@ Now, Metro runs in a single pass. Most of Metro's core transformations are run i
 - **[IR]**: Guard against identity mappings (T -> T) to prevent infinite recursion when remapping generic types.
 - **[IR]**: Fix directly providing a scoped `Map` instance not getting reused at injection sites.
 - **[IR]**: Fix graph extensions not being able to replace `@Binds`-provided bindings from parent graphs.
+- **[IR]**: Fix dynamic binding containers not being propagated to graph extensions in some cases.
 - **[IC]**: Fix an IC edge cases where generated assisted factory impl classes sometimes missed changes to injected constructor parameters in the target class.
 - **[FIR/IR/Reports]**: Restructure reports to use hierarchical nesting instead of top-level concatenated names. This fixes 'file name too long' exceptions when generating reports for deeply nested graphs. For example, the report file `reports/keys-populated-test_Graph_ChildGraph.txt` will now be generated as `reports/keys-populated/test/Graph/ChildGraph.txt`
 
@@ -185,10 +186,11 @@ Now, Metro runs in a single pass. Most of Metro's core transformations are run i
 
 Special thanks to the following contributors for contributing to this release!
 
-- [@inorichi](https://github.com/inorichi)
 - [@Egorand](https://github.com/Egorand)
 - [@JoelWilcox](https://github.com/JoelWilcox)
+- [@inorichi](https://github.com/inorichi)
 - [@japplin](https://github.com/japplin)
+- [@jonapoul](https://github.com/jonapoul)
 - [@vRallev](https://github.com/vRallev)
 
 0.10.4
--- a/compiler-tests/src/test/data/box/dependencygraph/dynamic/DynamicGraphReplacesExtensionBindingFromClass.kt
+++ b/compiler-tests/src/test/data/box/dependencygraph/dynamic/DynamicGraphReplacesExtensionBindingFromClass.kt
@@ -1,0 +1,33 @@
+// https://github.com/ZacSweers/metro/pull/1889
+// MODULE: app
+@DependencyGraph(AppScope::class)
+interface AppGraph {
+  val string: String
+  val subGraph: SubGraph
+}
+
+@GraphExtension
+interface SubGraph {
+  val string: String
+}
+
+@BindingContainer
+@ContributesTo(AppScope::class)
+object DefaultBinding {
+  @Provides fun string(): String = "abc"
+}
+
+// MODULE: test(app)
+@BindingContainer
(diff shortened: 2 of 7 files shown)


samples/android-app/src/main/kotlin/dev/zacsweers/metro/sample/android/MainActivity.kt
```diff
@@ -29,6 +29,8 @@ class MainActivity(private val fragmentFactory: FragmentFactory) :
     enableEdgeToEdge()
     super.onCreate(savedInstanceState)
 
+    // This is somehow the fake instance in tests, but ViewModel still sees the original binding
+    val tracker = (applicationContext as MetroApp).appGraph.tracker
```
rubensousa, line 33: So, for some reason, this during tests receives `FakeTracker`, but the ViewModel still sees `TrackerImpl`

samples/android-app/src/androidTest/kotlin/dev/zacsweers/metro/sample/android/MainActivityTest.kt
```diff
@@ -0,0 +1,38 @@
...
+class MainActivityTest {
+
+  private val tracker = FakeTrackerBindings.tracker
+
+  @Before
+  fun inject() {
+    val app =
+      InstrumentationRegistry.getInstrumentation().targetContext.applicationContext as MetroApp
+    val graph = createDynamicGraphFactory<AppGraph.Factory>(FakeTrackerBindings).create(app)
+    app.setTestGraph(graph)
```
rubensousa, line 24: Couldn't figure out another way to replace the graph. This isn't ideal, specially for the worker jobs that are launched immediately in `onCreate`. Those will still use the original graph

rubensousa:
> Sorry I do not want to add instrumentation tests. 

But why? I wanted to know how exactly we can write these tests with Metro, seems like it's the only missing link with Hilt. I was hoping you could either help with this sample or at least include a section in the docs with how to achieve this

ZacSweers:
They slow down CI quite a bit and are not necessary to demonstrate the createDynamicGraph API here.

rubensousa:
> They slow down CI quite a bit and are not necessary to demonstrate the createDynamicGraph API here.

Ok, fair point. But what about the docs then? I couldn't find any complete example that replaces dependencies. It would be great if you could share some hints. I can update the PR to just include a simple unit test then

ZacSweers:
> But what about the docs then? I couldn't find any complete example that replaces dependencies.

Not to sound glib but did you look at the docs for dynamic graphs? They have an example there: https://zacsweers.github.io/metro/latest/dependency-graphs/#dynamic-graphs

rubensousa:
> > But what about the docs then? I couldn't find any complete example that replaces dependencies.
> 
> Not to sound glib but did you look at the docs for dynamic graphs? They have an example there: https://zacsweers.github.io/metro/latest/dependency-graphs/#dynamic-graphs

Yes, I checked the docs. That example only works for plain unit tests. In instrumented tests, the graph is already created by the time the test starts since `Application.onCreate()` is called early, so you can't intercept the initial graph creation.

The changes I applied in this PR showcase the issue: even if you replace the graph before launching the activity, the ViewModel doesn't receive the fake instance. However, if you retrieve it from `appGraph` manually, it works. Do you have suggestions to address this? It's not clear for me, coming from Hilt, how to proceed here. Should I file an issue instead?

ZacSweers:
Ah!

> In instrumented tests, the graph is already created by the time the test starts since Application.onCreate() is called early, so you can't intercept the initial graph creation.

Sorry, this really isn't anything to do with metro then :). This is purely how you choose to weave your DI setup in your android app's lifecycle. Hilt does have some stuff for intercepting Application setup, but per the [FAQ](https://zacsweers.github.io/metro/latest/faq/#will-metro-add-support-for-hilt-features-or-hilt-interop) this library isn't really seeking to 1:1 replace/interop hilt APIs as many of them run through Hilt's custom Gradle bytecode transformations. There's nothing in Metro for "intercepting initial graph creation", that's something that's a layer of abstraction above metro.

I'm going to close this for now, I hope this makes sense.

samples/android-app/src/main/kotlin/dev/zacsweers/metro/sample/android/MetroApp.kt
```diff
@@ -22,10 +24,14 @@ class MetroApp : Application(), MetroApplication, Configuration.Provider {
...
   override fun onCreate() {
     super.onCreate()
-
     scheduleBackgroundWork()
   }
 
+  @VisibleForTesting
+  fun setTestGraph(graph: AppGraph) {
+    appGraph = graph
+  }
```
rubensousa, line 33: @ZacSweers this is what I mean here. Doing this replacement here isn't applying the changes for ViewModels somehow

rubensousa:
>  There's nothing in Metro for "intercepting initial graph creation"

But not only that, even replacing at runtime doesn't seem to work correctly. Even if I replace the `appGraph` being held in the application, looks like `ViewModel` instances do not see those changes. Is that intended? See here: https://github.com/ZacSweers/metro/pull/1893/changes#r2855020825

ZacSweers:
You're welcome to ask in the Q&A section of discussions for arch advice/debugging help, that's probably the best place for this rather than comments here

rubensousa:
Ok, opened a new thread there: https://github.com/ZacSweers/metro/discussions/1906

Issue #1887, opened 2026-02-24
Autogenerated assisted factory causes compiliation exception


hrach:
### Description

```kotlin
        extensions.configure<MetroPluginExtension> {
            generateAssistedFactories.set(true)
        }
```

code:
```kotlin
@AssistedInject
class RemoteConfigViewModel(
    private val remoteConfig: RemoteConfig,
    private val remoteKeyProviders: Set<RemoteKeyProvider>,
    dispatcher: CoroutineContext,
) : BaseViewModel() {
```

### Self-contained Reproducer

causes:
```kotlin
org.jetbrains.kotlin.fir.pipeline.IrGenerationExtensionException: Required a single abstract function for xxx.RemoteConfigViewModel.Factory but found multiple:
Required a single abstract function for xxxx.RemoteConfigViewModel.Factory but found multiple:
. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro. 
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.applyIrGenerationExtensions(convertToIr.kt:577)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.runActualizationPipeline(convertToIr.kt:279)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.convertToIrAndActualize(convertToIr.kt:150)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize(convertToIr.kt:115)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize$default(convertToIr.kt:88)
	at org.jetbrains.kotlin.cli.jvm.compiler.legacy.pipeline.JvmCompilerPipelineKt.convertToIrAndActualizeForJvm(jvmCompilerPipeline.kt:100)
	at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFir2IrPipelinePhase.executePhase(JvmFir2IrPipelinePhase.kt:30)
	at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFir2IrPipelinePhase.executePhase(JvmFir2IrPipelinePhase.kt:25)
```

### Metro version

0.10.4

### Previous working version

N/A

### IDE version

2026.1 EAP - IU-261.21525.39

### Context

Metro: 0.10.4
Kotlin: 2.3.10

hrach:
[...]

PR #1901, opened 2026-02-25
Work around IC bug with assisted factory generation

hrach:
Edit: Resolves ##1887

Hopefully a good repro & dummy fix for #1887.
Vibecoded.

>   - First build: Factory has create function, origin=GENERATED[AssistedFactoryClassDeclaration]
>  - IC build: Factory only has equals/hashCode/toString from Any, origin=IR_EXTERNAL_DECLARATION_STUB, declarations are Fir2IrLazySimpleFunction
>
>   The FIR callbacks aren't called for non-dirty files during IC. The fix needs to be at the IR level.

Could you please suggest which gradle task should I run to publish to maven local to try it out? Thx.

### DIFF
--- a/metrow
+++ b/metrow
@@ -91,7 +91,7 @@ case "$1" in
            fi
 
            echo "Publishing version $version locally..."
-           ./gradlew publishToMavenLocal -x dokkaGeneratePublicationHtml -PVERSION_NAME="${version}" --rerun-tasks
+           ./gradlew publishToMavenLocal -x dokkaGeneratePublicationHtml -PVERSION_NAME="${version}" -PRELEASE_SIGNING_ENABLED=false --rerun-tasks
        else
            echo "Publishing version $version..."
            ./gradlew publish -x dokkaGeneratePublicationHtml


Egorand:
> Could you please suggest which gradle task should I run to publish to maven local to try it out? Thx.

`./gradlew publishToMavenLocal -PRELEASE_SIGNING_ENABLED=false` is what worked for me recently.

hrach:
I tried the dummy fix and it works. This should help you design a proper fix. Thx a lot.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -805,6 +818,42 @@ internal fun IrClass.singleAbstractFunction(): IrSimpleFunction {
...
 }
 
+/**
+ * Under incremental compilation, auto-generated assisted factory interfaces loaded from cache may
+ * be missing their `create()` function because the FIR plugin callbacks aren't re-invoked for
+ * non-dirty files. This function reconstructs the missing abstract function from the parent class's
+ * `@AssistedInject` constructor parameters.
+ */
+context(context: IrMetroContext)
+private fun IrClass.repairAssistedFactoryAbstractFunction(): IrSimpleFunction? {
```
ZacSweers: This seems like, if that's the actual behavior, it would be a bug in kotlinc? I can't think of any scenario where a class would be generated but not its supertypes or methods

hrach: If you say so... I know nothing about FIR, IR, IC... to be sure.

ZacSweers: Found a simpler workaround

ZacSweers:
Found a simpler workaround and filed https://youtrack.jetbrains.com/issue/KT-84578. Thanks!

For future ref, please take a look at the CONTRIBUTING.md docs as it has a bunch of development instructions, including how to publish to maven local :)

hrach:
Thx! I did - but I haven't seen Contributing.md and the releasing.md was kinda... short :D 
I see it now.

image

ZacSweers:
yup it's in the tabs next to README
CleanShot 2026-02-25 at 18 13 26@2x

PR #1905, opened 2026-02-25
Fix optional root graph accessors

### LINKED ISSUE #1899: Nullable @OptionalBinding fails with error [Metro/MissingBinding]
### Description

We're using `@OptionalBinding` accessors a lot in our project to gate new features from production for instance. We'll always use `null` to indicate a missing binding.
This works in version `0.10.1` but breaks in `0.10.2+`.
If I replace the nullable version with a fake (non-null) version, it works.

The compile error I get
```
[Metro/MissingBinding] Cannot find an @Inject constructor or @Provides-annotated function/property for: my.package.Foo?
```

### Self-contained Reproducer

```kotlin
interface Foo
object FakeFoo : Foo

@DependencyGraph
interface Graph {
    @OptionalBinding
    val foo: Foo? get() = null // no working

    @OptionalBinding
    val fakeFoo: Foo get() = FakeFoo // working
}
```

### Metro version

0.11.0

### Previous working version

0.10.1

### IDE version

_No response_

### Context

_No response_

JonasAtAmo:
Resolves #1899

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ Changelog
 **Unreleased**
 --------------
 
+### New
+
+### Enhancements
+
+### Fixes
+
+- **[IR]**: Fix root graph accessors with `@OptionalBinding` accidentally reporting missing bindings.
+
+### Changes
+
+### Contributors
+
+Special thanks to the following contributors for contributing to this release!
+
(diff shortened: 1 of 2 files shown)


ZacSweers:
d'oh, this is actually an easy one to fix and specific to just when it's declared as a graph-level root accessor. Thanks for the repro

PR #1912, opened 2026-02-26
Add more diagnostic messaging in case the parent class is not resolvablee to an IrClass

### LINKED ISSUE #1883: Request: More helpful error messages when an upstream factory exposes a type that is not visible in the public API
### Description

I got an error today that was in the lines of:

```
 e: org.jetbrains.kotlin.fir.pipeline.IrGenerationExtensionException: Parent of this declaration is not a class: CLASS IR_EXTERNAL_DECLARATION_STUB OBJECT name:DataDatastoreBindings modality:FINAL visibility:public superTypes:[kotlin.Any]
```

It took me quite some time to find out what was wrong until I looked at metro source, and found [this] (https://github.com/ZacSweers/metro/blob/1366aede7e447a533af0c27d84a5b3bf99db4e63/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt#L1949-L1953) which helped me figure it out.

One of the dependencies I was providing in my :data:datastore module was not in my app module.

Is there a way to be able to get a more helpful error message? Even adding [this comment](https://github.com/ZacSweers/metro/blob/1366aede7e447a533af0c27d84a5b3bf99db4e63/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt#L1949-L1953) in the error message could be helpful already, because I fixed it 1 minute after seeing that comment.

### Alternatives/workarounds

_No response_

segunfamisa:
This PR adds some helpful diagnostic messaging during error reporting where the parent of a class is not resolvable to an IR class, and instead of showing an error like below, we show a more helpful error indicating that this may be happening when the declaration's parent is not resolvable to an IrClass

```txt
Caused by: java.lang.IllegalStateException: Parent of this declaration is not a class: 
CLASS IR_EXTERNAL_DECLARATION_STUB OBJECT name:DataDatastoreBindings modality:FINAL visibility:public superTypes:[kotlin.Any]
```

Closes #1883

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ Changelog
 
 Special thanks to the following contributors for contributing to this release!
 
-- [@JonasAtAmo](https://github.com/JonasAtAmo)
 - [@hrach](https://github.com/hrach)
+- [@JonasAtAmo](https://github.com/JonasAtAmo)
 - [@segunfamisa](https://github.com/segunfamisa)
 
 0.11.0


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/irDeclarationRendering.kt
```diff
@@ -60,6 +62,16 @@ internal fun IrDeclaration.humanReadableDiagnosticMetadata(): DiagnosticMetadata
...
+    if (parentClass.parent !is IrClass) {
+      context.reportCompat(
+        this,
+        MetroDiagnostics.METRO_ERROR,
+        "Could not read parent of '${parentClass.kotlinFqName}'. " +
+          "This may be a sign that it is used but not in the compile classpath.",
+      )
+
+      return@let
+    }
```
segunfamisa: I was not sure if I should add this check specifically for the `CallableMetadata` branch or for everything. I'm not sure which is preferable, but let me know if I should move this into the corresponding `if-block`

ZacSweers: I'm struggling to come up with a repro of any sort so I'm gonna just go with the safe option and move the string to the `metadata` string we're building for context here anyway

PR #1954, opened 2026-03-07
Pass creation extras to assistedMetroViewModel for ManualAssistedFactories

hvisser:
There are currently two variants for creating a `ViewModel` using `assistedViewModel`. The first variant passes `CreationExtras` to the `create` method of the factory. The only real use case for this IMHO is passing a `SavedStateHandle` although you can pass other parameters.

The second variant gives you a lamba to construct the `ViewModel` with a provided `ManualAssistedFactory`. It also sets up the default `CreationExtras`, but does not pass it to the lambda to create the view model using the provided factory. It looks like this is probably an oversight.

This PR passes the `CreationExtras` to the lamba, so that one can pull out any extras needed and pass it to the factory.

Ideally I think Metro should just be able to inject the `SavedStateHandle` by default, but I'm not sure how that fits in the current setup.

Example usage:

```kotlin
@AssistedInject
class MyViewModel(@Assisted private val myId: String, @Assisted private val savedState: SavedStateHandle): ViewModel() {
  @AssistedFactory
  @ContributesIntoMap(AppScope::class)
  @ManualViewModelAssistedFactoryKey(Factory::class)
  interface Factory: ManualViewModelAssistedFactory {
    fun create(myId: String, savedState: SavedStateHandle): MyViewModel
  }
}

@Composable
fun Test() {
  val vm: MyViewModel = assistedMetroViewModel<MyViewModel, MyViewModel.Factory>() {
    create("123", it.createSavedStateHandle())
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Changelog
 - **[FIR]**: Don't use a memoizing sequence for all FirSession instance as it seems that the IDE will mutate the underlying source lists in some cases.
 - **[FIR]**: Providers can now return instances of classes nested in the same container class.
 - **[IR]**: Fix codegen error when a scoped binding in a child graph supersedes the same-typed scoped binding from a parent graph and is used in a grandchild graph's multibinding. Basically, if graph A provides `Logger` and graph `B` also provides `Logger` (overriding `A`'s), graph `C` would incorrectly try to get it from `A` instead of `B`.
-- **[metrox-viewmodel-compose]**: Pass `CreationExtras` to the `createViewModel` lamba for `assistedMetroViewModel` when using `ManualViewModelAssistedFactory`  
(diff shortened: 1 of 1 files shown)


metrox-viewmodel-compose/src/commonMain/kotlin/dev/zacsweers/metrox/viewmodel/MetroViewModel.kt
```diff
@@ -72,7 +72,7 @@ public inline fun <
     } else {
       CreationExtras.Empty
     },
-  crossinline createViewModel: FactoryType.() -> VM,
+  crossinline createViewModel: FactoryType.(CreationExtras) -> VM,
```
yschimke, line 75: Yep, seems like an oversight, this should be passed on, so you can also do

```
            val savedStateHandle = extras.createSavedStateHandle()
```

hvisser:
Sorry, forgot to fetch first so rebased and fixed the conflict now.

PR #1980, opened 2026-03-13
Fix NoSuchFieldError when sharded graphs access @Includes dependency properties

vRallev:
In `GraphExpressionGenerator`, the `BoundInstance` and `GraphDependency` (getter case) handlers accessed @Includes graph dependency properties using `irGetProperty(irGet(thisReceiver), property)` directly. When code was generated inside a shard class, `thisReceiver` pointed to the shard's `this`, but the property lived on the main graph class. This generated bytecode like:

    this@Shard1.abc  // WRONG

instead of the correct:

    this@Shard1.graph.abc

At runtime this manifested as:

    java.lang.NoSuchFieldError: No field abc of type Abc in class MyGraphGraph$Impl$Shard1

The Kotlin/JVM backend generated a synthetic `access$` method on Shard1 to bridge the private field access from a lambda in the shard's init, but the field only existed on the main Impl class.

The fix routes all three call sites through `generatePropertyAccess()`, which already handles shard context correctly: for non-sharded properties (shardIndex == null) it generates `this.graph.property` instead of `this.property`.


vRallev:
> lgtm. Wanna add a changelog line?

Done!

PR #1983, opened 2026-03-13
Partial zensical support

ZacSweers:
See zensical-migration.md

CC @hossain-khan

### DIFF
--- a/docs/zensical-migration.md
+++ b/docs/zensical-migration.md
@@ -34,4 +34,5 @@ When Zensical adds versioning support:
 3. Update `scripts/delete_old_version_docs.sh` to use Zensical's equivalent
 4. Update `extra.version.provider` in `mkdocs.yml` if needed
 5. Check if the `site_url` `!ENV` workaround is still needed or if zensical handles the prefix natively
-6. Remove this file
+6. Ensure the `gh-pages` branch root files are preserved or migrated: `index.html` (redirects to `latest/`) and `404.html` (smart redirect for versioned docs). If Zensical provides native equivalents, migrate to those.
+7. Remove this file


scripts/run_local_docs_site.sh
```diff
@@ -1,17 +1,17 @@
 #!/bin/bash
 
-# The website is built using MkDocs with the Material theme.
-# https://squidfunk.github.io/mkdocs-material/
+# The website is built using Zensical (successor to MkDocs Material).
+# https://zensical.org/
```
hossain-khan, line 4: Yay! I like what I see! however, their splash screen spiral though!! 😵‍💫 😆

Image

scripts/run_local_docs_site.sh
```diff
@@ -1,17 +1,17 @@
...
 #
 # To run the site locally with hot-reload support, use:
 # ./scripts/run_local_docs_site.sh
 
-# Check if mkdocs is installed
-if ! command -v mkdocs &> /dev/null; then
-    echo "mkdocs is not installed. Please run:"
+# Check if zensical is installed
+if ! command -v zensical &> /dev/null; then
+    echo "zensical is not installed. Please run:"
```
hossain-khan, line 14: Wait, I thought `zensical` is rust based. Is running pip enough to have this tool ready?

mkdocs.yml
```diff
@@ -3,10 +3,11 @@
 # source .venv/bin/activate
 # pip install -r .github/workflows/mkdocs-requirements.txt
 # ./scripts/copy_docs_files.sh
-# mkdocs serve
+# zensical serve
 #
 # ℹ️ NOTE: Deployment is done via `mike` for versioned site using GitHub Actions
 #          For manual deployment, use `./scripts/deploy_metro_docs_site.sh`
+# ℹ️ NOTE: See docs/zensical-migration.md for details on the partial Zensical migration
```
hossain-khan, line 10: 👍🏽 

Did you notice improvement over mkdocs on hotreload? 

I also wonder if their search indexing and other under the hood tools will also be more efficient once we are able to get away from mkdocs. 🤔  I will read up on this new tool 📖

docs/zensical-migration.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Completing the Migration
+
+When Zensical adds versioning support:
+
+1. Remove `mkdocs`, `mkdocs-material`, `mkdocs-material-extensions`, `mike`, and their transitive dependencies from `.github/workflows/mkdocs-requirements.txt`
+2. Update `mike deploy` calls in CI workflows (`.github/workflows/docs-site.yml`, `docs-site-manual.yml`) and `scripts/deploy_metro_docs_site.sh` to use Zensical's versioning
+3. Update `scripts/delete_old_version_docs.sh` to use Zensical's equivalent
+4. Update `extra.version.provider` in `mkdocs.yml` if needed
+5. Remove this file
```
hossain-khan: Worth adding one step - to ensure that following two files stays intact or is migrated to adapt for Zensical, in case they natively support equivalent files.
- https://github.com/ZacSweers/metro/blob/gh-pages/404.html
- https://github.com/ZacSweers/metro/blob/gh-pages/index.html

hossain-khan (approved):
Changes look good overall. Some thoughts!

scripts/run_local_docs_site.sh
ZacSweers: apparently

mkdocs.yml
ZacSweers: I didn't test, hot reload doesn't seem to work anymore since the mike integration :/

hossain-khan: I tested it out locally, hot reload worked for me using zensical.

PR #1987, opened 2026-03-14
[DOCS] Python version requirement and venv setup guide

hossain-khan:
## Summary
Added some note on local docsite run setup - had some issues when trying it out on my new macbook.

---

Zensical (https://pypi.org/project/zensical/) requires python 3.10+

For some reason latest macos comes with 3.9, so added minor note.

Also, added the best practice for using `.venv` for setting this up.

### DIFF
--- a/scripts/run_local_docs_site.sh
+++ b/scripts/run_local_docs_site.sh
@@ -2,12 +2,17 @@
 
 # The website is built using Zensical (successor to MkDocs Material).
 # https://zensical.org/
-# It requires Python to run.
+# It requires Python >=3.10 to run.
+#
+# To set up the environment, create a virtual environment and activate it (first time only):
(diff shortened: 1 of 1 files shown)


scripts/run_local_docs_site.sh
```diff
@@ -2,12 +2,17 @@
 
 # The website is built using Zensical (successor to MkDocs Material).
 # https://zensical.org/
-# It requires Python to run.
+# It requires Python >=3.10 to run.
+#
+# To set up the environment, create a virtual environment and activate it (first time only):
+#   python3 -m venv .venv
+#   source .venv/bin/activate
+#
```
ZacSweers, line 10: I don't want to be prescriptive on setting up virtual envs

PR #1992, opened 2026-03-15
Fix binding supersession in multi-level graph extensions 

### LINKED ISSUE #1974: Replacement using @Binds does not apply in child graph extensions
### Description

Similar to #1885, It appears that if you have a DependencyGraph `AppGraph` with some type `Logger`, replace it in a child graph `UserGraph` with an `@Binds` method, and when a graph `FeatureGraph` (as a child of `UserGraph`) injects `Logger`,  it'll always receive the original binding from `AppGraph`, not the replacement from `UserGraph`.

This does not occur if the replacement in `UserGraph` is made with `@Provides` or `@ContributesBinding`

### Self-contained Reproducer

```kotlin
interface Logger {
  val id: String
}

class BaseLogger(override val id: String) : Logger

abstract class UserScope private constructor()
abstract class FeatureScope private constructor()

interface AppGraph {
  val userGraph: UserGraph
}

@GraphExtension(UserScope::class)
interface UserGraph {
  val feature: FeatureGraph
  val logger: Logger
}

interface AppLoggerBindings {
  @Provides
  fun appLogger() : Logger = BaseLogger("app")
}

@ContributesTo(UserScope::class)
interface UserLoggerBindings {
  @Provides
  fun userLogger() : BaseLogger = BaseLogger("user")

  @Binds
  val BaseLogger.bindAsLogger: Logger
}

@GraphExtension(FeatureScope::class)
interface FeatureGraph {
  val logger: Logger // id is "app" when it should be "user"
}
```

### Metro version

0.11.1

### Previous working version

n/a

### IDE version

_No response_

### Context

_No response_

jonamireh:
## Summary
Fixes #1974 by skipping keys that have already been collected in their parent


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/graph/BindingGraphGenerator.kt
```diff
@@ -699,7 +699,11 @@ internal class BindingGraphGenerator(
       for ((key, factories) in extendedNode.providerFactories) {
         // Dynamic parent bindings take precedence over child's directly provided keys
         val isDynamicInParent = isDynamicParent && key in extendedNode.dynamicTypeKeys
-        if (key !in node.directlyProvidedKeys || isDynamicInParent) {
+        if (
+          (key !in node.directlyProvidedKeys || isDynamicInParent) &&
```
ZacSweers: We need to preserve the `isDynamicInParent` escape hatch here and below, as this would otherwise overrule dynamic keys in dynamic graphs

ZacSweers:
Had some tweaks locally that I had ready to push up!

PR #2025, opened 2026-03-20
Clean up deprecated native targets

ZacSweers:


LionZXY:
@ZacSweers I’m sorry to bother you 👉👈. I just wanted to check whether it’s possible to reintroduce support for macosX64?

At the moment, this is the only blocker we face in migrating from Kotlin Inject. 

Although the Kotlin team already considers these targets to be deprecated, dropping them for Metro makes it **impossible to use Metro for apps that have already been published on the App Store**. This is because the App Store requires a macosX64 target for apps that previously supported it.

> Apps for Macs with Apple silicon. If your Mac app requires the high performance of Apple silicon, you can make your app available on the Mac App Store only to Macs with an M1 chip or later. To only support Macs with Apple silicon, your app must require a minimum OS version of macOS Monterey 12 or later, **and have never supported Intel-based Macs**. In Xcode, update the Architectures build setting for your target and specify to build ARM64 only. Apps can add Intel support and lower the minimum OS version in an update but, once released, they cannot return to only supporting Apple silicon.

Source: https://developer.apple.com/app-store/submitting/

Is there anything I can do about this? Perhaps I could make a contribution, or take full responsibility for supporting the target with a guaranteed SLA?

ZacSweers:
I'm open to a PR to add them back (but continuing to note they are deprecated with JB's deprecation) given the context [here](https://kotlinlang.slack.com/archives/C0KLZSCHF/p1775552300894569?thread_ts=1775051562.095799&cid=C0KLZSCHF).

Issue #2021, opened 2026-03-19
createDynamicGraphFactory compiler crash when @GraphExtension depends on replaced scoped binding


grandstaish:
### Description

This bug is a bit tricky for me to explain, I don't fully understand it and it requires quite a specific setup (which our app happens to have). 

Setup:
- an app graph with a scoped instance (e.g. a singleton http client)
- graph extension with a field-injection function (e.g. for injecting an activity)
- the class being injected must depend on that scoped dependency from the app graph
- a test using `createDynamicGraphFactory` that overrides the scoped dependency

When compiling the test code, the compiler crashes. 

### Self-contained Reproducer

[metro-dynamic-graph-repro.zip](https://github.com/user-attachments/files/26126948/metro-dynamic-graph-repro.zip)

```base
./gradlew run  

> Task :run
real

BUILD SUCCESSFUL in 820ms
```

```bash
./gradlew test

> Task :compileTestKotlin FAILED
e: org.jetbrains.kotlin.fir.pipeline.IrGenerationExtensionException: Could not find a reportable source graph declaration for DynamicAppGraphImpl_Z8V9gDs.ActivityGraphImpl. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro. 
        at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.applyIrGenerationExtensions(convertToIr.kt:577)
        at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.runActualizationPipeline(convertToIr.kt:279)
        at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.convertToIrAndActualize(convertToIr.kt:150)
        at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize(convertToIr.kt:115)
        at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize$default(convertToIr.kt:88)
        at org.jetbrains.kotlin.cli.jvm.compiler.legacy.pipeline.JvmCompilerPipelineKt.convertToIrAndActualizeForJvm(jvmCompilerPipeline.kt:100)
        at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFir2IrPipelinePhase.executePhase(JvmFir2IrPipelinePhase.kt:30)
[...]

PR #2032, opened 2026-03-21
Fix compiler crash when replacing scoped bindings in dynamic graph

grandstaish:
Fix for https://github.com/ZacSweers/metro/issues/2021

Note: this fix was found using AI. I figure just setting up the test and branch is helpful anyway, even if the solution isn't quite right. Please feel free to change anything you want or close this if it's not helpful!


ZacSweers (commented):
While this covers the test case, I think this is functionally patching a symptom rather than the underlying issue. Basically, what's happening here is that we're omitting both the original `HttpClient` and the dynamic one for different reasons, leaving us missing both in the end.

I have an idea I'm gonna try on your branch

ZacSweers:
Got it working, thanks for the repro!

PR #2052, opened 2026-03-27
Excludes generated data class members from accessor candidates

kevinguitar:
One of our devs ran into a compiler crash by changing a class to a data class with the `@Includes` usage:

```
Caused by: java.lang.IllegalStateException: Must have at least two locations to report duplicate bindings. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro. 
	at dev.zacsweers.metro.compiler.graph.MutableBindingGraph.reportDuplicateBindings(BindingGraph.kt:767)
	at dev.zacsweers.metro.compiler.graph.MutableBindingGraph.reportDuplicateBindings(BindingGraph.kt:446)
	at dev.zacsweers.metro.compiler.ir.graph.IrBindingGraph.reportDuplicateBindings(IrBindingGraph.kt:433)
	at dev.zacsweers.metro.compiler.ir.graph.IrBindingGraph.realGraph$lambda$1$0(IrBindingGraph.kt:110)
	at dev.zacsweers.metro.compiler.ir.graph.BindingLookup.lookup$compiler(BindingLookup.kt:627)
	at dev.zacsweers.metro.compiler.ir.graph.IrBindingGraph.realGraph$lambda$1(IrBindingGraph.kt:109)
	at dev.zacsweers.metro.compiler.graph.MutableBindingGraph.populateGraph(BindingGraph.kt:206)
	at dev.zacsweers.metro.compiler.graph.MutableBindingGraph.seal(BindingGraph.kt:130)
```

After debugging, I found out that the duplicate binding is pointing to a generated data class member 👇 

CleanShot 2026-03-27 at 16 17 29

The fix is to exclude those generated members from accessor candidates.

_Disclosure: Fix was made with the help of AI_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ class HomeFactory(...) : BaseFactory<HomeFactory>
 
 - Update shaded `androidx.tracing` to 2.0.0-alpha04.
 
+### Contributors
+
+Special thanks to the following contributors for contributing to this release!
+
+- [@KevinGuitar](https://github.com/KevinGuitar)
+
 0.12.0
 ------
 


ZacSweers:
hmm, I think we'll wanna find a more general solution for this. We already have logic for excluding inherited functions from `Any`, my thinking is it makes sense to exclude these (component and copy) functions too. Thoughts?

kevinguitar:
It sounds good to me! I will revisit this PR today 🙇

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -537,6 +538,9 @@ internal fun IrClass.allCallableMembers(
     .letIf(excludeAnyFunctions) {
       it.filterNot { function -> function.isInheritedFromAny(context.irBuiltIns) }
     }
+    .letIf(excludeGeneratedDataClassMembers) {
+      it.filterNot { function -> function.origin == Origins.FirstParty.GENERATED_DATA_CLASS_MEMBER }
```
kevinguitar: I double checked this check, it covers copy as well

Image

kevinguitar: Let me know if you want to join the `GENERATED_DATA_CLASS_MEMBER` check in existing `isInheritedFromAny`, I wasn't quite sure if it's the right place to put it

ZacSweers: the origin checks won't work across modules so we shouldn't leverage that

kevinguitar: Got it, let me add a test to check this 👍

kevinguitar:
Made one more attempt, please review 🙇

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -1566,6 +1569,17 @@ internal fun IrFunction.isToStringOnAny(): Boolean {
     hasShape(dispatchReceiver = true, regularParameters = 0)
 }
 
+internal fun IrFunction.isCopyOnAny(): Boolean {
```
ZacSweers: nit: these names aren't quite accurate, this isn't a function from `Any`

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -1566,6 +1569,17 @@ internal fun IrFunction.isToStringOnAny(): Boolean {
...
+  return name == StandardNames.DATA_CLASS_COPY
+}
+
+internal fun IrFunction.isComponentNOnAny(): Boolean {
+  return name.asString().startsWith(StandardNames.DATA_CLASS_COMPONENT_PREFIX) &&
+    hasShape(dispatchReceiver = true, regularParameters = 0) &&
```
ZacSweers: nice

compiler-tests/src/test/data/box/dependencygraph/IncludesDataClass.kt
```diff
@@ -0,0 +1,22 @@
+// Repro for a bug where using @Includes on a class is fine, but not a data class.
+data class ExternalDependencies(
```
ZacSweers, line 2: we should have a non-data class with a no-arg copy() function here that does get picked up

Possibly also have a non-data class with component functions that we still ignore

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
```diff
@@ -1543,8 +1543,11 @@ internal val IrFunction.regularParameters: List<IrValueParameter>
     return parameters.filter { it.kind == IrParameterKind.Regular }
   }
 
-internal fun IrFunction.isInheritedFromAny(irBuiltIns: IrBuiltIns): Boolean {
-  return isEqualsOnAny(irBuiltIns) || isHashCodeOnAny() || isToStringOnAny()
+internal fun IrFunction.isInheritedFromAny(irBuiltIns: IrBuiltIns, isDataClass: Boolean): Boolean {
+  return isEqualsOnAny(irBuiltIns) ||
+    isHashCodeOnAny() ||
+    isToStringOnAny() ||
+    (isDataClass && (isCopyOnAny() || isComponentNOnAny()))
```
ZacSweers: thoughts on if we should always ignore component functions?

ZacSweers (commented):
small nit I'll fix before merging

compiler-tests/src/test/data/box/dependencygraph/IncludesDataClass.kt
kevinguitar: Ohh right, nice scenarios!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/ir/ir.kt
kevinguitar: I think so, I cannot think of a single case where we need a component function to be an accessor, after all, those functions duplicate all getters

PR #2058, opened 2026-03-28
Update kdoc for implicit map key usages

kevinguitar:


metrox-viewmodel/README.md
```diff
@@ -103,7 +103,7 @@ class CustomViewModel(@Assisted val param1: String, @Assisted val param2: Int) :
   // ...
 
   @AssistedFactory
-  @ManualViewModelAssistedFactoryKey(Factory::class)
+  @ManualViewModelAssistedFactoryKey
```
ZacSweers, line 106: from your other PR, this should probably be fixed to point at the returned `CustomViewModel`?

kevinguitar: No, `@ManualViewModelAssistedFactoryKey` requires a factory class, so implicit class key will work here 👍  My fix was for `@ViewModelAssistedFactoryKey`

https://github.com/ZacSweers/metro/blob/8237df278ffd8de793b52cb6e084de1312b120b3/metrox-viewmodel/src/commonMain/kotlin/dev/zacsweers/metrox/viewmodel/ViewModelKeys.kt#L53-L55

PR #2059, opened 2026-03-28
Remove implicitClassKey from ViewModelAssistedFactoryKey

kevinguitar:
I'm trying to integrate metrox-viewmodel in my project, and noticed `@ViewModelAssistedFactoryKey` has an implicit class key. From my understanding, we will always need to specify the class in this annotation because Factory is a separate instance from the ViewModel:

```kotlin
@AssistedInject
class SettingsViewModel(@Assisted val userId: String) : ViewModel() {

  @AssistedFactory
  @ViewModelAssistedFactoryKey(SettingsViewModel::class) // <--- we always need to specify VM class
  @ContributesIntoMap(AppScope::class)
  fun interface Factory : ViewModelAssistedFactory
}
```

Not sure if this was deliberate or not, please close the PR if it's not relevant 🙇


ZacSweers:
yup good catch, I think this makes sense

PR #2072, opened 2026-03-30
Honor provider-holder origins in FIR replacement merging

vRallev:
With `generateContributionProviders` enabled, binding contributions are represented by generated provider-holder containers that only carry `@Origin`. FIR replacement scanning only inspected the generated container itself, so `@ContributesBinding(replaces = ...)` on the origin class was ignored.

Scan provider-holder origins while collecting native replacements in `ContributedInterfaceSupertypeGenerator` so replaced classes still remove extension-generated contributions during FIR merging. This restores the expected replacement behavior for the remaining metro-extensions regressions under contribution providers.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
@@ -464,16 +464,34 @@ internal class ContributedInterfaceSupertypeGenerator(
 
     // Process replacements from native contributions
     val unmatchedReplacements = mutableSetOf<ClassId>()
-    contributionClassLikes
-      .mapNotNull {
-        val symbol = it.toClassSymbol(session)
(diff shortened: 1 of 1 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -464,16 +464,34 @@ internal class ContributedInterfaceSupertypeGenerator(
...
+              symbol?.getContainingClassSymbol()
+            }
+          }
+        )
+
+        // When generateContributionProviders is enabled, binding contributions are represented by
+        // provider-holder binding containers that only carry @Origin. Scan the origin class too so
+        // its original @ContributesBinding(replaces = ...) annotations still participate in FIR
+        // merging, matching the IR fallback logic.
+        for ((containerClassId, isBindingContainer) in contributionMappingsByClassId) {
```
ZacSweers, line 486: should we gate this on that option?

PR #2096, opened 2026-04-07
Fix support for mapkeys and qualifiers on explicit bindings

kevinguitar:
Noticed an issue with `generateContributionProviders` enabled.

```
error: found an @IntoMap annotation without any @MapKey annotations. This may happen if this is an external declaration that has a map key annotation that is not visible to this compilation. Please check the original source.
```

Btw, I was thinking maybe it could be useful to mirror some test cases related to binding/ multibindings for `generateContributionProviders` (like fast init), since it's quite a fundamental change, and based on my understanding, it's handled quite differently with the option disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ Changelog
 
 - Mark generated Circuit factories as `@Deprecated(HIDDEN)` + disable them in IDE as they're not necessary there.
 
+### Contributors
+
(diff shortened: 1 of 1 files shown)


ZacSweers:
There is a dedicated contribution providers box variant if you put it in the same subdir as the rest of them

kevinguitar:
ohh nice! I completely missed that 🙇

PR #2097, opened 2026-04-07
Fix StackOverflowError during IDE indexing with generateContributionProviders

### LINKED ISSUE #2095: StackOverflowError during IDE indexing when generateContributionProviders is enabled
### Description

I noticed after enabling `generateContributionProviders` option, our project cannot be indexed, and it seems like it's caused by a bunch of StackOverflowError (we observed 13 of them happened simultaneously).

I consolidated them into two stack traces, you can find them here:
https://gist.github.com/kevinguitar/8ad017225e54ecbb5b285638d14b7647

### Self-contained Reproducer

No reproducer for now, but let me know if I can help provide more info in any form 🙇 

### Metro version

0.13.2

### Previous working version

N/A

### IDE version

_No response_

### Context

_No response_

Sultan1993:
Fixes #2095

`getTopLevelClassIds()` in `ContributionsFirGenerator` calls `isAnnotatedWithAny()` and `usesContributionProviderPath()`, which trigger annotation resolution that re-enters `FirExtensionDeclarationsSymbolProvider`'s lazy cache init in the IDE, causing infinite recursion.

Defers the filtering to `generateTopLevelClassLikeDeclaration()`, which runs after the cache is initialized. `getTopLevelClassIds()` now returns a superset using only predicate-based queries.

### DIFF
--- a/ide-integration-tests/test-project/build.gradle.kts
+++ b/ide-integration-tests/test-project/build.gradle.kts
@@ -12,4 +12,5 @@ kotlin { jvmToolchain(21) }
 metro {
   generateAssistedFactories.set(true)
   enableTopLevelFunctionInjection.set(true)
+  generateContributionProviders.set(true)
 }
--- a/ide-integration-tests/test-project/src/main/kotlin/TestSources.kt
+++ b/ide-integration-tests/test-project/src/main/kotlin/TestSources.kt
@@ -55,6 +55,17 @@ interface EnumQualifierGraph {
   @Provides @By(ByKey.Two) fun provideTwo(): Int = 2
 }
 
+// Contribution provider: exercises getTopLevelClassIds() during IDE indexing
+interface Greeter {
+  fun greet(): String
+}
+
+@ContributesBinding(AppScope::class)
+@Inject
+class GreeterImpl : Greeter {
+  override fun greet(): String = "hello"
+}
+
 // Viewing generated supertypes
 @ContributesTo(AppScope::class)
 interface Base {


Sultan1993:
Not sure why the workflows are not running. @ZacSweers could you help?

ZacSweers:
Because you haven't contributed to this repo before, I have to manually approve them.

Was this change written by AI?

Sultan1993:
> Because you haven't contributed to this repo before, I have to manually approve them.
> 
> Was this change written by AI?

Yes, the implementation was AI-assisted. I identified the contribution aggregation codepath from the stack trace based on my experience with Anvil at Square, and directed the AI toward the fix. The FIR re-entrancy specifics were the AI's contribution

ZacSweers:
Ok. AI use is fine but for future reference the COC of this project asks that you disclose this upfront and confirm you have cleaned it up before submitting. Both the description and comment in code are quite tedious, please de-fluff it a bit. Thanks!

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributionsFirGenerator.kt
```diff
@@ -226,7 +221,13 @@ internal class ContributionsFirGenerator(session: FirSession, compatContext: Com
 
   @ExperimentalTopLevelDeclarationsGenerationApi
   override fun generateTopLevelClassLikeDeclaration(classId: ClassId): FirClassLikeSymbol<*>? {
-    if (classId !in topLevelContributionHolders) return null
+    val holder = topLevelContributionHolders[classId] ?: return null
+
+    // Deferred from getTopLevelClassIds() to avoid IDE reentrancy.
+    if (!holder.contributingClassSymbol.usesContributionProviderPath(session)) {
+      topLevelContributionHolders.remove(classId)
```
ZacSweers: this double mutation is a bit awkward and kind of typical of an AI assistant racing toward the simple solution rather than a good one. I'd rather us just correctly handle skipping any not in a contribution path at usages, as these callbacks can run multiple times.

Sultan1993: Created a helper method that checks `usesContributionProviderPath()` at usage sites

ZacSweers:
Needs a test

Sultan1993:
> Did you confirm the new test fails without your fix? This is another thing I am wary up with AI-generated PRs where the test comes later 😓

I tried publishing locally and running the test, but I couldn't reproduce the issue. The StackOverflow error requires a re-entrancy cycle through the FirExtensionDeclarationsSymbolProvider's lazy classNamesInPackageCache.

@kevinguitar Would you be able to test this fix against your project?

I believe the test is still good to have as enabling `generateContributionProviders` in the IDE test project ensures this code path is exercised during IDE analysis going forward, catching future regressions as the test project grows.

ZacSweers:
Another alternative here may be to just disable this feature in IDE mode all together, since these classes aren't intended to be user-visible anyway (see #2100 for an example)

kevinguitar:
I can test against this branch today 👌

kevinguitar:
Looks like the fix works for our project, no longer seeing StackOverflowError anymore. 🎉  However, I see one more, but this one seems easy to fix because it happens during typing (seems like). I'll send a separate PR for this
https://gist.github.com/kevinguitar/4088a13d18973c1ff724f90a4b35160d

ZacSweers:
I'm ok with accepting this on anecdotal tests but def would prefer repro tests first in the future. Thanks all!

PR #2103, opened 2026-04-08
Wrap type resolution in InjectedClassFirGenerator with try-catch

### LINKED ISSUE #2104: IDE error: Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl
### Description

Encountered an error (full stacktrace [here](https://gist.github.com/kevinguitar/4088a13d18973c1ff724f90a4b35160d)) in the IDE while editing the code.
```
org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.getConeType(FirTypeUtils.kt:201)
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.toConeTypeProjection(FirTypeUtils.kt:161)
	at org.jetbrains.kotlin.fir.resolve.providers.impl.FirTypeResolverImpl.resolveUserType(FirTypeResolverImpl.kt:195)
	at org.jetbrains.kotlin.fir.resolve.providers.impl.FirTypeResolverImpl.resolveType(FirTypeResolverImpl.kt:394)
	at dev.zacsweers.metro.compiler.fir.MetroFirTypeResolver$LocalMetroFirTypeResolver.resolveType(TypeResolverFactory.kt:132)
	at dev.zacsweers.metro.compiler.fir.generators.InjectedClassFirGenerator$InjectedClass.parentClassHasMemberInjections(InjectedClassFirGenerator.kt:271)
	at dev.zacsweers.metro.compiler.fir.generators.InjectedClassFirGenerator.generateNestedClassLikeDeclaration(InjectedClassFirGenerator.kt:506)
	at dev.zacsweers.metro.compiler.fir.generators.CompositeMetroFirDeclarationGenerationExtension.generateNestedClassLikeDeclaration(CompositeMetroFirDeclarationGenerationExtension.kt:149)
	at dev.zacsweers.metro.compiler.fir.generators.KotlinOnlyFirGenerationExtension.generateNestedClassLikeDeclaration(KotlinOnlyFirGenerationExtension.kt:46)

```

### Self-contained Reproducer

Not able to repro in test

### Metro version

0.13.2 + cherry picking the fix from https://github.com/ZacSweers/metro/pull/2097

### Previous working version

N/A

### IDE version

Android Studio Panda 3 | 2025.3.3

### Context

_No response_

kevinguitar:
Encountered an error (full stacktrace [here](https://gist.github.com/kevinguitar/4088a13d18973c1ff724f90a4b35160d)) in the IDE while editing the code.
```
org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.getConeType(FirTypeUtils.kt:201)
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.toConeTypeProjection(FirTypeUtils.kt:161)
	at org.jetbrains.kotlin.fir.resolve.providers.impl.FirTypeResolverImpl.resolveUserType(FirTypeResolverImpl.kt:195)
	at org.jetbrains.kotlin.fir.resolve.providers.impl.FirTypeResolverImpl.resolveType(FirTypeResolverImpl.kt:394)
	at dev.zacsweers.metro.compiler.fir.MetroFirTypeResolver$LocalMetroFirTypeResolver.resolveType(TypeResolverFactory.kt:132)
	at dev.zacsweers.metro.compiler.fir.generators.InjectedClassFirGenerator$InjectedClass.parentClassHasMemberInjections(InjectedClassFirGenerator.kt:271)
	at dev.zacsweers.metro.compiler.fir.generators.InjectedClassFirGenerator.generateNestedClassLikeDeclaration(InjectedClassFirGenerator.kt:506)
	at dev.zacsweers.metro.compiler.fir.generators.CompositeMetroFirDeclarationGenerationExtension.generateNestedClassLikeDeclaration(CompositeMetroFirDeclarationGenerationExtension.kt:149)
	at dev.zacsweers.metro.compiler.fir.generators.KotlinOnlyFirGenerationExtension.generateNestedClassLikeDeclaration(KotlinOnlyFirGenerationExtension.kt:46)

```

I tried the fix locally, and I can confirm the error goes away for me.

Resolves #2104

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/InjectedClassFirGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/InjectedClassFirGenerator.kt
@@ -271,7 +271,8 @@ internal class InjectedClassFirGenerator(session: FirSession, compatContext: Com
               try {
                 resolver.resolveType(it)
               } catch (_: Exception) {
-                // Generic type resolution may fail during IDE indexing
+                // Generic type resolution may fail during IDE indexing because
+                // this isn't a deep resolve
                 return@any false
               }
             }


kevinguitar:
I see, could you please point me to the best place to reproduce such issues 🙏

ZacSweers:
Should be able to in compiler-tests, it might be a supertype that is an impl dep within the module but not visible to a consuming one. We can still mapNotNull but we should try to resolve it rather than only discard it

kevinguitar:
Tried a few ways, but looks like I cannot reproduce it 🤔  I will close the PR for now and file an issue instead, sorry for the trouble and thanks for the review!

ZacSweers:
These three bits

```
org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.getConeType(FirTypeUtils.kt:201)
	at org.jetbrains.kotlin.fir.types.FirTypeUtilsKt.toConeTypeProjection(FirTypeUtils.kt:161)
```

Suggest it's specifically a generic type's projection. I think this one?

CleanShot 2026-04-08 at 14 30 08@2x

which in turn points here

CleanShot 2026-04-08 at 14 31 05@2x

I suspect you have a generic type arg that is unresolved somewhere. So your filter wouldn't have addressed this anyway I think since it's not the raw type itself. We'd need to so this resolving ourselves since kotlin isn't doing it here. Tbh I am not sure how to do that safely, so I may just wrap this in a try/catch for now and call it a day.

kevinguitar:
Oh, it could be. Though the files I navigated through weren't using generics, I don't know how IDE indexes files, if it aggressively analyzes all dependent files (which I think is the case), or files in the same module, that could be it.

> so I may just wrap this in a try/catch for now and call it a day.

I can do that and verify the fix in my project, but not really sure about creating a repro test. Any idea?

ZacSweers:
if you have a consistent repro, it'd be helpful if you could try/catch and print the type it's trying to use. if you hit `fir.render()` it'll render whatever it can see, that should help find the specific type that's causing this and help get a repro

kevinguitar:
Got it, let me spend some time on this today 👍

kevinguitar:
It is indeed a generic type 👍  I was able to locate the exact type and replicate the structure in a test, but the test is passing fine. I'll reopen this PR with a try-catch for now.
```kotlin
// MODULE: lib0
interface Atomic

// MODULE: lib1(lib0)
interface Base<T, R> : Atomic

// MODULE: lib2(lib0, lib1)
abstract class Base2<T> : Base<String?, T>

// MODULE: main(lib0, lib1, lib2)
data class Foo(val a: Int)

@Inject
class Example : Base2<List<Foo>?>()

interface Bar

@Inject
@ContributesBinding(AppScope::class)
class BarImpl(example: Example) : Bar

interface AppGraph {
  val bar: Bar
}

fun box(): String {
  return "OK"
}

```

kevinguitar:
Also, looks like the error exists only during IDE indexing, I don't see it during compilation so far

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/InjectedClassFirGenerator.kt
```diff
@@ -268,7 +268,12 @@ internal class InjectedClassFirGenerator(session: FirSession, compatContext: Com
             if (it is FirResolvedTypeRef) {
               it.coneType
             } else {
-              resolver.resolveType(it)
+              try {
+                resolver.resolveType(it)
+              } catch (_: Exception) {
+                // Generic type resolution may fail during IDE indexing
```
ZacSweers:
```suggestion
                // Generic type resolution may fail during IDE indexing because
                // this isn't a deep resolve
```

PR #2117, opened 2026-04-09
Add macosX64 target

LionZXY:
Although the Kotlin team already considers these targets to be deprecated, **dropping them for Metro makes it impossible to use Metro for apps that have already been published on the App Store**. This is because the App Store requires a macosX64 target for apps that previously supported it.

> Apps for Macs with Apple silicon. If your Mac app requires the high performance of Apple silicon, you can make your app available on the Mac App Store only to Macs with an M1 chip or later. To only support Macs with Apple silicon, your app must require a minimum OS version of macOS Monterey 12 or later, **and have never supported Intel-based Macs**. In Xcode, update the Architectures build setting for your target and specify to build ARM64 only. Apps can add Intel support and lower the minimum OS version in an update but, once released, they cannot return to only supporting Apple silicon.

Source: https://developer.apple.com/app-store/submitting/

That’s why I think macosX64 is worth enabling

References:
- Kotlin Slack discussion [here](https://kotlinlang.slack.com/archives/C0KLZSCHF/p1775552300894569?thread_ts=1775051562.095799&cid=C0KLZSCHF)
- YouTrack issue [here](https://youtrack.jetbrains.com/issue/KT-78660#focus=Comments-27-13603171.0-0)
- Clean up PR [here](https://github.com/ZacSweers/metro/pull/2025)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,14 @@ Changelog
 ### Changes
 
 - Mark generated Circuit factories as `@Deprecated(HIDDEN)` + disable them in IDE as they're not necessary there.
+- Add back deprecated `macosX64()`, `tvosX64()`, and `watchosX64()` targets for now due to [KT-78660 (comment)](https://youtrack.jetbrains.com/issue/KT-78660#focus=Comments-27-13603171.0-0).
 
 ### Contributors
 
 Special thanks to the following contributors for contributing to this release!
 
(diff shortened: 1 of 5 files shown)


ZacSweers (approved):

ZacSweers:
Can you add a line to the changelog?

LionZXY:
> Can you add a line to the changelog?

I apologize for not responding in time, thank you very much for the quick feedback loop

PR #2131, opened 2026-04-11
Fix default values getting overwritten

asapha:
Hey, I got a subtle one…
If you have multiple default values of the same type in a constructor, they all end up having the same value when injected.

The test results in

```
Expected <Example(foo=foo, one=1, bar=bar, two=2)>, actual <Example(foo=bar, one=2, bar=bar, two=2)>.
```

Dumped via `scratch.kt`

```kotlin
  class MetroFactory : Factory<Example> {
      […]
       override operator fun invoke(): Example {
      return Companion.newInstance(foo = <this>.#bar.invoke(), one = <this>.#two.invoke(), bar = <this>.#bar.invoke(), two = <this>.#two.invoke())
    }
  }
```

Can repro in 0.12


ZacSweers (commented):
Thanks for the repro, though this is an unconventional thing to do

ZacSweers:
Sidenote - "Can repro in 0.12" isn't super helpful considering that's several versions behind. In the future it'd be helpful to know if you tried latest

PR #2134, opened 2026-04-13
Don't support inject-less circuitinject classes

ZacSweers:


compiler-tests/src/test/data/box/api/circuit/PresenterClassNoInjectionWithCircuitParams.kt
ZacSweers, line 1:
```suggestion
// ENABLE_CIRCUIT
// COMPILER_VERSION: 2.3
```

PR #2146, opened 2026-04-15
Fix redundant class key in metrox-android/README.md ActivityKey example

Daiji256:
The explicit class key in the README example is redundant and triggers a warning.

This removes the unnecessary argument and uses the implicit key instead.


ZacSweers (approved):
Thanks!

PR #2159, opened 2026-04-17
Fix typo in circuit.md

anddani:
This change adds a missing `>` to `Set<Presenter.Factory>` in the Circuit Integration documentation.


ZacSweers (approved):
Thanks!

PR #2198, opened 2026-04-24
Avoid duplicate contributed graph supertypes

vRallev:
Metro now promotes original `@ContributesTo` interfaces as direct graph supertypes for ObjC/Swift export visibility. That broke graphs that already explicitly extend the same contributed interface, because FIR saw the promoted interface and the source-declared interface as duplicate direct supertypes.

This changes FIR supertype generation to track a graph's existing direct supertypes and skip promoting a contributed parent when it is already declared.

### DIFF
--- a/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
+++ b/compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
@@ -264,7 +264,11 @@ internal class ContributedInterfaceSupertypeGenerator(
   ): List<ConeKotlinType> {
     // For generated @DependencyGraph classes (from external FIR extensions), FIR calls this
     // method instead of computeAdditionalSupertypes. Delegate to the shared implementation.
-    return computeContributionSupertypes(klass, typeResolver)
+    return computeContributionSupertypes(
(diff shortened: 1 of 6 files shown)


compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/generators/ContributedInterfaceSupertypeGenerator.kt
```diff
@@ -264,20 +264,30 @@ internal class ContributedInterfaceSupertypeGenerator(
...
     classLikeDeclaration: FirClassLikeDeclaration,
     resolvedSupertypes: List<FirResolvedTypeRef>,
     typeResolver: TypeResolveService,
   ): List<ConeKotlinType> {
-    return computeContributionSupertypes(classLikeDeclaration, typeResolver)
+    return computeContributionSupertypes(
+      classLikeDeclaration = classLikeDeclaration,
+      typeResolver = typeResolver,
+      existingSupertypeClassIds =
+        resolvedSupertypes.mapNotNullTo(mutableSetOf()) { it.coneType.classId },
```
ZacSweers: nit: I have a `mapNotNullToSet` helper

vRallev: Done

ZacSweers (approved):

PR #2203, opened 2026-04-25
Misc tracing and graph gen optimizations

ZacSweers:


ZacSweers:
Some claude-generated notes for posterty

---

- Fine-grained perfetto tracing across `Build GraphNode`, `Build binding graph`, `[Impl] Generate graph`/`Process shards`, `BindingContainer lookup`/`Iterate declarations`, `MembersInjector`, `Find supertype binding containers`, and `Core transformers` (per-transformer + per-class names) — every previously-opaque "black box" phase is now broken down into named sub-spans.
- Hoist the `typeKey in bindings` fast-path check above `stack.withEntry(...)` in `populateGraph` to skip per-dependency `Entry` allocation + push/pop in the common case.
- Gate `IrBindingStack` push/pop/copy/init logger work on `logger.type != Type.None` so `entry.toString()` (which renders a table) and string interpolation aren't paid when the logger is `NONE`.
- Make `IrBindingStack.Entry.graphContext` lazy — expensive parent-traversal/buildString work in `injectedAt`/`providedAt`/`requestedAt`/`generatedExtensionAt`/`memberInjectedAt` only runs when the stack is actually rendered.
- Memoize `nameHint`, `dependencies` (where non-trivial), and `reportableDeclaration` on `IrBinding` subtypes that previously did eager `rawType()`/`buildString`/`capitalizeUS`/`resolveSourceDeclaration` work at construction.
- Short-circuit `BindingContainerTransformer.findContainer` and the supertype-binding-container loop for platform types (`kotlin.*`, `java.*`, `android.*`, `androidx.*`, `kotlinx.*`, `scala.*`) — they can never be containers.
- Cache failure paths in `loadExternalBindingContainer` (missing Dagger factory class, missing metadata) so repeated lookups don't re-report the same diagnostic; switch all `BindingContainerTransformer` cache writes from `=` to `putIfAbsent` for correct concurrent behaviour.
- Replace wasteful full-`metroAnnotations()` classifier calls in `Process declarations` with targeted `isAnnotatedWithAny`/`qualifierAnnotation()` lookups, and reuse the already-classified `annotations` from the outer loop instead of re-running it on the same declaration.
- Add a `log(() -> String)` lazy-message overload on `IrMetroContext` and use it in the per-class `CoreTransformers.visitClassInner` log so `kotlinFqName` isn't traversed when `logFile` is unset.
- Move the `declaration.isLocal` skip above `trace("Visit X")` in `ContributionIrTransformer` so local classes don't allocate a trace span and name string just to bail.

PR #2217, opened 2026-04-28
Add a test for generating providers in a generated dependency graph

kevinguitar:
This PR adds a compiler test for generating providers in a generated dependency graph. 
~This PR supports FIR-generated providers to be processed in a generated dependency graph.~

Previously, when you added a provider in a generated graph:
```kotlin
@DependencyGraph
interface FeatureGraph {
  @Provides
  fun provideParam(): String = "foo"
}
```
the compilation would fail with this error:
```
No expected factory class generated for /MyActivity.FeatureGraph.provideParam. Report this bug with a repro case at https://github.com/zacsweers/metro/issues/new. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro.
org.jetbrains.kotlin.fir.pipeline.IrGenerationExtensionException: No expected factory class generated for /MyActivity.FeatureGraph.provideParam. Report this bug with a repro case at https://github.com/zacsweers/metro/issues/new. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro.
```
**Edit**: Annotated the dependency graph with `@IROnlyFactories` resolves the issue.

~and this is because when a `@Provides` function is generated by a FIR extension, it's generated at a phase after Metro's `ProvidesFactoryFirGenerator` has already looked at the `classSymbol.declarationSymbols` (is this expected?!) to find provider functions and generate corresponding factory classes.~

~Generating the factory in IR phase (`BindingContainerTransformer`) resolves the issue.~

### DIFF
--- a/compiler-tests/src/test/data/box/api/GeneratedProvidesInGraph.kt
+++ b/compiler-tests/src/test/data/box/api/GeneratedProvidesInGraph.kt
@@ -14,6 +14,7 @@ interface StringComponent {
 class Application
 
 // The extension generates a nested interface inside Application:
+//   @IROnlyFactories
 //   @DependencyGraph(AppScope::class)
 //   interface AppGraph {
 //     @Provides
--- a/compiler-tests/src/test/data/box/api/GeneratedProvidesInGraph.kt
+++ b/compiler-tests/src/test/data/box/api/GeneratedProvidesInGraph.kt
@@ -1,0 +1,29 @@
+package test
+
+// Custom annotation that triggers the test extension to generate a @DependencyGraph
+// with a provider inside the annotated class
+@Target(AnnotationTarget.CLASS)
+annotation class GenerateProvidesInGraph
+
+@ContributesTo(AppScope::class)
+interface StringComponent {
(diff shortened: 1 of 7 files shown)


ZacSweers:
> ProvidesFactoryFirGenerator has already looked at the classSymbol.declarationSymbols (is this expected?!)

remember that the way the FIR phases work is not FirExtension -> another FirExtension. It runs all extensions' callbacks at once, then moves to the next phase and runs all their callbacks again.

In general I'm not a fan of this solution as it just tries to punt the problem out of FIR for all generation, but that creates a whole set of other potential issues. I'd carved out the `@IROnlyFactories` bit to allow extensions to denote their providers to use this, don't want all providers to use this. Your extension I think just needs to use that annotation like the circuit extension demonstrates.

kevinguitar:
Got it 👍  Let me use `@IROnlyFactories` in the test case and revert the functional part of changes, thanks!

ZacSweers (approved):

Issue #2216, opened 2026-04-27
docs: prevent docs site redirect loop in the future


hossain-khan:
## Summary
Follow up from following issues
* https://github.com/ZacSweers/metro/issues/2215
* https://github.com/ZacSweers/metro/issues/1216

I think fallback to latest snapshot is a good option if this happens again. If snapshot is also gone, then fallback to last known stable.

* https://github.com/ZacSweers/metro/blob/gh-pages/versions.json

👀 I will take a look into it soon.

hossain-khan:
<img width="797" height="80" alt="Image" src="https://github.com/user-attachments/assets/1c6a6e79-5b84-40b1-9738-df33fda73c75" />

Found the problematic commit https://github.com/ZacSweers/metro/commit/2508c7cb1f17cfe47445f00c40059c4e00557702

hossain-khan:
## 🕵🏽  Findings

**Bug: `delete_old_version_docs.sh` deletes the final release when a pre-release exists in the same series**

**What happened:**
When `1.0.0` was tagged and the docs workflow ran, it:
1. Deployed `1.0.0` as `[latest]` via `mike deploy`
2. Immediately ran `delete_old_version_docs.sh` to clean up old versions
3. The script found both `1.0.0` and `1.0.0-RC4` in the `1.0` series
4. Used `sort -V | tail -n1` to pick the "latest" — but `sort -V` ranks `1.0.0-RC4` **higher** than `1.0.0` (non-numeric suffixes sort after bare versions)
5. Decided `1.0.0-RC4` was the latest 1.0.x release and **deleted `1.0.0`** — including the `latest` alias — just seconds after deploying it

**Root cause:**
`delete_old_version_docs.sh` only treats `-SNAPSHOT` versions as pre-releases. Versions with `-RC*`, `-alpha`, `-beta`, etc. are bucketed alongside final releases, and `sort -V` incorrectly ranks them as newer than the final release.

**Relevant log evidence (run #1265):**
- Before deployment: `mike list` showed `1.0.0-RC4 [latest]`
- Deployed: `mike deploy --update-aliases --push 1.0.0 latest` ✅
- Cleanup script ran and logged: `Cleaning up old versioned site - deleting 1.0.0` ❌

**Fix needed:**
[...]

PR #2218, opened 2026-04-28
Fix delete_old_version_docs.sh incorrectly deleting stable releases when pre-releases exist

hossain-khan:
## Summary
Fixes re-direct loop caused by current script always deleting **`1.0.0`** release

Details
* https://github.com/ZacSweers/metro/issues/2216

----
sort -V ranks pre-release suffixes (e.g. -RC4) higher than the bare stable version (e.g. 1.0.0), causing the cleanup script to delete the freshly-deployed stable release and keep the old RC instead.

Split the version classification into three buckets:
- snapshots   (*-SNAPSHOT): keep only the single latest globally
- releases    (X.Y.Z):      keep latest per X.Y series (unchanged)
- prereleases (*-RC*, etc.): keep latest per X.Y series ONLY when no
                             stable release exists for that series yet;
                             deleted entirely once a stable release ships

### DIFF
--- a/scripts/delete_old_version_docs.sh
+++ b/scripts/delete_old_version_docs.sh
@@ -2,7 +2,8 @@
 
 # This script cleans up old versioned mkdocs site by keeping:
 # 1. Only the LATEST snapshot version (not one per major.minor series)
-# 2. The latest patch version for each major.minor series (for releases)
+# 2. The latest patch version for each major.minor series (for stable releases)
+# 3. The latest pre-release (RC/alpha/beta) per series, ONLY if no stable release exists yet
 #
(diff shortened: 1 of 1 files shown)


hossain-khan:
FYI: Redirect loop is back because of the bug, I will be merging it, and then initiate manual docs release for 1.0.0 again

ZacSweers (approved):

hossain-khan:
First time seeing this! LOL

Screenshot 2026-04-28 at 12 03 53 AM copy

ZacSweers:
¯\_(ツ)_/¯

PR #2226, opened 2026-04-29
Allow MetroFirDeclarationGenerationExtension.enableFirInIde to be overridden

kevinguitar:
Realized that `enableFirInIde` is not opened, but I'm not sure if it's not ready to be used yet, or it was a mistake. Be free to close it if it's irrelevant. 🙇


ZacSweers (approved):
Oversight! Forgot it was an abstract class and not an interface

PR #2249, opened 2026-05-05
Fix FirNamedArgumentExpression crash in renderAnnotationArgument

### LINKED ISSUE #2244: FirNamedArgumentExpression not handled in renderAnnotationArgument
### Description

Metro crashes when an annotation uses named arguments with class references:

When in renderAnnotationArgument() (fir.kt) doesn't handle FirNamedArgumentExpression, falling through to the else branch which calls reportCompilerBug("Unexpected annotation argument type")

Potential Fix: Add a case that unwraps to the inner expression:
```kotlin
is FirNamedArgumentExpression -> {
    renderAnnotationArgument(session, arg.expression, typeResolver)
}
```

### Self-contained Reproducer

Repro test: `compiler-tests/src/test/data/box/interop/dagger/NamedAnnotationArgumentClassRef.kt`
```kotlin
import com.squareup.anvil.annotations.ContributesBinding
import com.squareup.anvil.annotations.MergeComponent
import com.squareup.anvil.annotations.optional.SingleIn
import javax.inject.Inject
import javax.inject.Qualifier

@Qualifier
@Retention(AnnotationRetention.RUNTIME)
annotation class MyQualifier(val value: kotlin.reflect.KClass<*>)

interface Service

@ContributesBinding(AppScope::class)
class ServiceImpl @Inject constructor() : Service

// This annotation uses a named argument (value = ...) which triggers
// FirNamedArgumentExpression in the FIR tree.
@MyQualifier(value = Service::class)
@ContributesBinding(AppScope::class)
class QualifiedService @Inject constructor() : Service

@SingleIn(AppScope::class)
@MergeComponent(AppScope::class)
interface AppGraph {
  val service: Service
}

fun box(): String {
  assertNotNull(graph.service)
  return "OK"
}
```

### Metro version

1.0.0

### Previous working version

N/A

### Kotlin version

2.2.21

### IDE version

_No response_

### Platform

_No response_

### Context

_No response_

aitorvs:
Fixes #2244
                                                                                                                                                                                                   
`renderAnnotationArgument()` doesn't handle `FirNamedArgumentExpression`, causing a crash when annotations use named arguments with class references (e.g. `@MyQualifier(value = Service::class)`).


aitorvs:
@ZacSweers we x-posted :)

Feel free to close this one if it's a dup now.

compiler/src/main/kotlin/dev/zacsweers/metro/compiler/fir/fir.kt
```diff
@@ -482,6 +482,10 @@ private fun renderAnnotationArgument(
         arg.arguments.map { renderAnnotationArgument(session, it, typeResolver) }
       }
 
+      is FirNamedArgumentExpression -> {
+        renderAnnotationArgument(session, arg.expression, typeResolver)
+      }
+
```
ZacSweers:
```suggestion
```

gradle/libs.versions.toml
```diff
@@ -24,7 +24,7 @@ jetbrains-lifecycle-navigation3 = "2.10.0"
 jetbrains-navigation3 = "1.1.0"
 dagger = "2.59.2"
 gjf = "1.35.0"
-jdk = "24"
+jdk = "25"
```
ZacSweers:
```suggestion
jdk = "24"
```

ZacSweers:
Interesting, it seems specific to compiled (java?) annotations maybe

aitorvs:
> Interesting, it seems specific to compiled (java?) annotations maybe

I hit this with one of our KSP processors, that uses KotlinPoet to emit `@SomeAnnotation(value = SomeClass::class)` with the named `value = ` syntax that produced the crash.

PR #2266, opened 2026-05-11
Fix member injection `Invalid type arg` error when class has more generic type parameters than its parent

VirtualParticle:
Right now, the build fails when a class has more generic type parameters than its parent and both have member injections (this includes a non-generic parent):

```kotlin
@HasMemberInjections
abstract class Parent<T> {
    @Inject lateinit var foo: String
}

@Inject
class Child<T, R> : Parent<T>() {
    @Inject lateinit var bar: List<Int>
}
```

The reason for this is that, in this example, `Child`'s type paremeters are being passed in to `irInvoke` and are then compared with the type parameters on `Parent`'s inject function:

```kotlin
internal fun IrBuilderWithScope.irInvoke(
  typeArgs: List<IrType>? = null,
  // ...
) {
  // ...
  typeArgs?.let { // `typeArgs` is the Child's types
    for ((i, typeArg) in typeArgs.withIndex()) {
      if (i >= call.typeArguments.size) { // `call` uses the Parent's types
        reportCompilerBug(
          "Invalid type arg $typeArg at index $i for callee ${callee.owner.dumpKotlinLike()}"
        )
      }
      call.typeArguments[i] = typeArg
    }
  }
}
```

Since `Child` has more type parameters than `Parent`, the comparison fails and we get an error:

```
Invalid type arg org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeOnlyClassifierImpl@a2ddaad8 at index 1 for callee @JvmStatic
@JsStatic
fun <T : Any?> injectFoo(@Assisted instance: Parent<T>, foo: String) {
  return instance.#foo = foo
}

. This is possibly a bug in the Metro compiler, please report it with details and/or a reproducer to https://github.com/zacsweers/metro. 
org.jetbrains.kotlin.fir.pipeline.IrGenerationExtensionException: Invalid type arg org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeOnlyClassifierImpl@a2ddaad8 at index 1 for callee @JvmStatic
@JsStatic
fun <T : Any?> injectFoo(@Assisted instance: Parent<T>, foo: String) {
  return instance.#foo = foo
}
```

The fix for this is to pass in the types of the function rather than the child class. I tried this out locally in my use case and it seems to work fine.

### DIFF
--- a/compiler-tests/src/test/data/box/inject/member/GenericMemberInjectionWithFewerParentTypeParameters.kt
+++ b/compiler-tests/src/test/data/box/inject/member/GenericMemberInjectionWithFewerParentTypeParameters.kt
@@ -1,0 +1,53 @@
+@HasMemberInjections
+abstract class GrandParent {
+    @Inject lateinit var grandParentString: String
+}
+
+@HasMemberInjections
+abstract class Parent<T : Any> : GrandParent() {
+    @Inject lateinit var parentT: T
+}
+
+@Inject
+class ExampleClass<T : Any, R : Any> : Parent<R>() {
+  @Inject lateinit var value: T
+  @Inject lateinit var values: List<T>
+  @Inject lateinit var mapValues: Map<T, List<T>>
+
+  // Setter mid-properties to ensure ordering doesn't matter
+  lateinit var functionSet: T
+
+  @Inject
+  fun functionMemberInject(value: T) {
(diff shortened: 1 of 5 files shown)


ZacSweers:
Thanks!

PR #2307, opened 2026-05-14
Add a test for generated graph extensions

kevinguitar:
I added a test to cover the generated graph extension use case using Metro's FIR API.

Fyi, I noticed an error during commit, causing the pre-commit hook to be skipped. Installed manually and committed again helped.
```
Warning: kempt not found on PATH. Install with `brew install ZacSweers/tap/kempt-fmt` or `cargo install kempt-fmt`. Skipping pre-commit format.
```

### DIFF
--- a/compiler-tests/src/test/kotlin/dev/zacsweers/metro/compiler/api/GenerateGraphExtensionExtension.kt
+++ b/compiler-tests/src/test/kotlin/dev/zacsweers/metro/compiler/api/GenerateGraphExtensionExtension.kt
@@ -1,0 +1,417 @@
+// Copyright (C) 2025 Zac Sweers
+// SPDX-License-Identifier: Apache-2.0
+package dev.zacsweers.metro.compiler.api
+
+import dev.zacsweers.metro.compiler.MetroOptions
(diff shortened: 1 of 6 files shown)


compiler-tests/src/test/kotlin/dev/zacsweers/metro/compiler/api/GenerateGraphExtensionExtension.kt
```diff
@@ -0,0 +1,417 @@
...
+ *     @ContributesTo(AppScope::class)
+ *     @GraphExtension.Factory
+ *     fun interface Factory {
+ *       fun create(@Provides text: String): LoginGraph
+ *     }
+ *   }
+ * }
+ * ```
+ */
+internal class GenerateGraphExtensionExtension(session: FirSession) :
```
kevinguitar, line 88: There is a decent amount of code duplication in test extensions. I can offer some refactoring to extract common patterns to a util class in the same module. Let me know what you think 🙏

ZacSweers (approved):
Nice

PR #2364, opened 2026-05-30
Use provider holders for binding containers

tcmulcahy:
This is all AI-generated code and I don't even understand it myself yet. Please treat it as raw AI agent output that hasn't been properly vetted by a human yet. That includes the following description:

Pure binding contributions represented as binding containers still created per-origin contribution mirror metadata that large graphs had to resolve and retain during compilation. On large graphs this could keep the Kotlin daemon near its heap limit and cause heavy full-GC pressure.

Route those pure binding contributions through generated provider-holder containers instead. Keep concrete factory generation unless explicit contribution providers are enabled, so direct injections of implementation classes continue to work. Also shorten generated provider names to avoid filesystem filename limits in deeply nested graphs.


ZacSweers:
Closing this out for now, can revisit as needed in the future

PR #2391, opened 2026-06-04
Add bounded backoff to the native Lock

ZacSweers:


runtime/src/nonWebTest/kotlin/dev/zacsweers/metro/internal/LockTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+@OptIn(ExperimentalAtomicApi::class)
+class LockTest {
+  @Test
+  fun reentrant() {
```
swankjesse, line 21: Do you really want reentrant? I think reentrant is bad code hygiene!

(if you rely on a lock being reentrant, it means there’s somewhere in your program that doesn’t know if it’s being called with a lock or not, and I think this is a bad idea)

runtime/src/nonWebTest/kotlin/dev/zacsweers/metro/internal/LockTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    val state = AtomicInt(0)
+    val release = AtomicInt(0)
+
+    val holder =
+      async(Dispatchers.Default) {
+        lock.lock()
+        try {
+          state.incrementAndFetch()
+          while (release.load() == 0) {
+            // Keep the lock held by this worker thread until the test releases it.
```
swankjesse, line 45: I think it would be conventional to call `yield()` or something here

runtime/src/nativeMain/kotlin/dev/zacsweers/metro/internal/Lock.native.kt
```diff
@@ -8,45 +10,56 @@ import kotlin.concurrent.atomics.decrementAndFetch
...
 private object CurrentThread {
-  val id = Any()
+  val id = monotonicThreadIdCounter.incrementAndFetch()
 }
 
-@OptIn(ExperimentalNativeApi::class, ExperimentalAtomicApi::class)
+private const val SPINS_BEFORE_SLEEP = 64
+private const val MAX_SLEEP_MICROS = 1_000u
+
 internal actual class Lock {
```
swankjesse, line 25: It is conventional to call this thing `SpinLock` because `Lock` gives the caller the expectation that this is a reasonable thing that’ll use queues and not eat CPU.

SpinLocks are bad in general. They have their uses but you probably don‘t want ’em.

runtime/src/nativeMain/kotlin/dev/zacsweers/metro/internal/Lock.native.kt
```diff
@@ -8,45 +10,56 @@ import kotlin.concurrent.atomics.decrementAndFetch
...
-  private val locker_ = AtomicInt(0)
-  private val reenterCount_ = AtomicInt(0)
+  private val locker = AtomicInt(0)
```
swankjesse, line 26: This needs documentation

PR #2424, opened 2026-06-10
Fix typo in Circuit integration documentation

shenghaiyang:
Changed the duplicated `Set<Presenter.Factory>` to `Set<Ui.Factory>`


ZacSweers (approved):
Thanks!

PR #2437, opened 2026-06-11
metro contribution container check

tcmulcahy:
In progress


tcmulcahy:
I think https://github.com/ZacSweers/metro/pull/2434 is a better approach.

PR #2441, opened 2026-06-12
Update compiler API doc.

ychescale9:
Add missing `MetroContributionHintExtension` in compiler API doc after https://github.com/ZacSweers/metro/pull/2407.


ZacSweers (approved):
Thanks! CC @kevinguitar

PR #2442, opened 2026-06-12
Fully qualify scope names for hint functions

BraisGabin:
This is a failing test for an issue that I found in my project.

What we discover is that we can reproduce it when:
- We have two different modules
- Each module has a different scope but with the same name (e.g. `MyScope`)
- Each module has a `@DependecyGraph` using the scope of its module
- One module depends on the other
- The module that is depended on defines at least one `@ContributeBinding` (if it defines more than one a new error arises)
- `generateContributionProviders` is active

In my project this throws 3 errors:

```
e: file:///double-scope/src/main/kotlin/FeatureGraph.kt:7:1 An interface cannot extend a class.
e: file:///double-scope/src/main/kotlin/FeatureGraph.kt:7:1 Only one class can appear in a supertype list.
e: file:///double-scope/src/main/kotlin/FeatureGraph.kt:7:1 Cannot extend an object.
```

But it looks like the test only shows the one of them.

I also know how to "workaround" some of them:
- If I change `FeatureGraph` from `interface` to `abstract class` the first error goes away.
- If I remove one of the `@ContributesBinding` the second error goes away.


ZacSweers:
heh, you hit an old TODO. Multiple scopes with the same name are rare so it just hasn't come up before. Will fix but you can work around it for now by using differently named scopes, which I think is the better move in the long run anyway.

BraisGabin:
> Will fix but you can work around it for now by using differently named scopes, which I think is the better move in the long run anyway.

Fully agree and that's what we did and we will not revert it even if this is fixed. But it was a fun bug to track and we thought it was worth to report it because we expens some time to figure out which was the problem.

ZacSweers:
Yup thanks for the report!

PR #2449, opened 2026-06-16
Fix KMP output collisions in MetroArtifactCopyTask

FletchMcKee:
First I just want to say that this is an amazing framework. I can't imagine how much thought and effort went into this, and the analysis reports are awesome. I've been integrating Metro into a KMP project and the only issue we've had is the Android reports wouldn't generate which I didn't think warranted creating an issue. However I believe I found a simple fix for this.

The main cause of missing analysis reports in a KMP project is the targets would collide in MetroArtifactCopyTask. 
MetroGradleSubplugin was correctly doing this:
```kotlin
val reportsDir =
  extension.reportsDestination.map { baseDir ->
    // Include target name to avoid collisions in KMP projects where multiple targets
    // may have compilations with the same name (e.g., both jvm and android have "main")
    // Use chained dir() calls instead of joining with "/" to avoid Windows path issues
    listOf(kotlinCompilation.target.name, kotlinCompilation.name)
      .filter(String::isNotBlank)
      .fold(baseDir) { dir, segment -> dir.dir(segment) }
  }
```

But MetroArtifactCopyTask wasn't repeating the same pattern:
```kotlin
task.reportsDir.set(
  project.layout.buildDirectory.dir("tmp/metro/reporting/${sourceCompilation.name}")
)
```

Replicating the pattern fixed this issue and I could revert the `reportsDestination.isPresent` changes if desired. That was my first hunch since we apply Metro in a convention plugin and alter whether `reportsDestination` is set based on a few things. But I think this will also help improve the consistency of these tasks since they're no longer registered in an `afterEvaluate` block and the `isPresent` read is deferred until task execution. The tasks are always registered, but they're skipped if `reportsDestination.isPresent` is false.

I added some test cases but also tested it locally on FieldSpottr, and without the change it would yield the same results I had seen in our project:

Before:
```console
Project: :shared
Graphs analyzed: 1

Graph: dev.zacsweers.fieldspottr.di.IosFSGraph
  Total bindings: 48
  Scoped: 10, Unscoped: 38
  Avg dependencies: 1.31
  Binding types: {Alias=19, BoundInstance=1, ConstructorInjected=21, Multibinding=2, Provided=5}
  Longest path: 5 nodes
    PermitRepository -> PermitRepositoryImpl -> Logger -> FSGraph -> IosFSGraph
  Highest fan-in: PermitRepository (5 dependents)

> Task :shared:generateMetroGraphHtml
Generating Metro graph visualizations from file:///.../FieldSpottr/shared/build/reports/metro/graphMetadata.json
Including analysis data from file:///.../FieldSpottr/shared/build/reports/metro/analysis.json
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/dev-zacsweers-fieldspottr-di-IosFSGraph.html
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/index.html
```
After:
```console
Project: :shared
Graphs analyzed: 3

Graph: dev.zacsweers.fieldspottr.di.AndroidFSGraph
  Total bindings: 49
  Scoped: 9, Unscoped: 40
  Avg dependencies: 1.35
  Binding types: {Alias=19, BoundInstance=2, ConstructorInjected=21, Multibinding=2, Provided=5}
  Longest path: 7 nodes
    PermitRepository -> PermitRepositoryImpl -> FSAppDirs -> ContextFSAppDirs -> FileSystem -> FSGraph -> AndroidFSGraph
  Highest fan-in: PermitRepository (5 dependents)

Graph: dev.zacsweers.fieldspottr.di.IosFSGraph
  Total bindings: 48
  Scoped: 10, Unscoped: 38
  Avg dependencies: 1.31
  Binding types: {Alias=19, BoundInstance=1, ConstructorInjected=21, Multibinding=2, Provided=5}
  Longest path: 5 nodes
    PermitRepository -> PermitRepositoryImpl -> Logger -> FSGraph -> IosFSGraph
  Highest fan-in: PermitRepository (5 dependents)

Graph: dev.zacsweers.fieldspottr.di.JvmFSGraph
  Total bindings: 48
  Scoped: 10, Unscoped: 38
  Avg dependencies: 1.35
  Binding types: {Alias=19, BoundInstance=1, ConstructorInjected=21, Multibinding=2, Provided=5}
  Longest path: 6 nodes
    PermitRepository -> PermitRepositoryImpl -> SqlDriverFactory -> JvmSqlDriverFactory -> FSAppDirs -> DesktopFSAppDirs
  Highest fan-in: PermitRepository (5 dependents)

> Task :shared:generateMetroGraphHtml
Generating Metro graph visualizations from file:///.../FieldSpottr/shared/build/reports/metro/graphMetadata.json
Including analysis data from file:///.../FieldSpottr/shared/build/reports/metro/analysis.json
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/dev-zacsweers-fieldspottr-di-AndroidFSGraph.html
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/dev-zacsweers-fieldspottr-di-IosFSGraph.html
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/dev-zacsweers-fieldspottr-di-JvmFSGraph.html
Generated file:///.../FieldSpottr/shared/build/reports/metro/html/index.html
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,8 +89,8 @@ Example rich output (note that in rich terminals this would have color and marku
 Special thanks to the following contributors for contributing to this release!
 
 - [@BraisGabin](https://github.com/BraisGabin)
-- [@ychescale9](https://github.com/ychescale9)
 - [@FletchMcKee](https://github.com/FletchMcKee)
+- [@ychescale9](https://github.com/ychescale9)
 
 ### [Consider sponsoring Metro's development](https://www.zacsweers.dev/sponsoring-metro/)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Example rich output (note that in rich terminals this would have color and marku
 - **[IR/JS]** Fix `Map<K, () -> V>` multibindings accessed through provider-style map factories on Kotlin/JS. Generated maps now store callable function values instead of Metro `Provider` objects.
 - **[IR/KLIB]** Fix generated `@Binds` implementations on KLIB backends. Metro now emits concrete identity bodies for inherited `@Binds` members where JS, Native, and Wasm validate abstract members during deserialization.
 - **[IR/KLIB]** Keep generated graph and shard backing fields private while preserving generated access through properties. This avoids backing-field visibility validation failures on KLIB backends.
+- **[gradle]** Fix multiplatform output collisions in MetroArtifactCopyTask.
 
 ### Changes
 
@@ -89,6 +90,7 @@ Special thanks to the following contributors for contributing to this release!
 
 - [@BraisGabin](https://github.com/BraisGabin)
 - [@ychescale9](https://github.com/ychescale9)
+- [@FletchMcKee](https://github.com/FletchMcKee)
 
 ### [Consider sponsoring Metro's development](https://www.zacsweers.dev/sponsoring-metro/)
 
--- a/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroGradleSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroGradleSubplugin.kt
@@ -288,21 +288,18 @@ public class MetroGradleSubplugin @Inject constructor(problems: Problems) :
           .fold(baseDir) { dir, segment -> dir.dir(segment) }
       }
 
-    val artifactsTask = MetroArtifactCopyTask.register(project, reportsDir, kotlinCompilation)
+    if (extension.reportsDestination.isPresent) {
+      val artifactsTask = MetroArtifactCopyTask.register(project, reportsDir, kotlinCompilation)
 
-    artifactsTask.configure { task ->
-      task.onlyIf("reportsDestination is present") { extension.reportsDestination.isPresent }
-    }
-
-    project.tasks.withType(GenerateGraphMetadataTask::class.java).configureEach { task ->
-      task.onlyIf("reportsDestination is present") { extension.reportsDestination.isPresent }
-      task.projectPath.set(project.path)
-      task.compilationName.set(kotlinCompilation.name)
-      task.graphJsonFiles.from(
-        artifactsTask
-          .flatMap { it.reportsDir.dir("graph-metadata") }
-          .map { it.asFileTree.matching { it.include("*.json") } }
-      )
+      project.tasks.withType(GenerateGraphMetadataTask::class.java).configureEach { task ->
+        task.projectPath.set(project.path)
+        task.compilationName.set(kotlinCompilation.name)
+        task.graphJsonFiles.from(
+          artifactsTask
+            .flatMap { it.reportsDir.dir("graph-metadata") }
+            .map { it.asFileTree.matching { it.include("*.json") } }
+        )
+      }
     }
 
     val metroOptions =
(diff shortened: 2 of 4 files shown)


ZacSweers:
Nice find. Looks like there's a couple failures. Also can you add a mention to the CHANGELOG.md?

ZacSweers:
What was wrong with the JS/wasm testing?

FletchMcKee:
I thought it might have been failing due to issues with JS after seeing test failures like this in the [reports](https://github.com/user-attachments/files/29059619/_mKqGRfObKs.html):
```
> Task :compileKotlinJs FAILED
e: Kotlin/JS does not support generating top-level declarations with incremental compilation enabled. See https://youtrack.jetbrains.com/issue/KT-82395 and https://youtrack.jetbrains.com/issue/KT-82989. Either disable enableTopLevelFunctionInjection, generateContributionHints, generateContributionHintsInFir for JS targets or disable JS IC.
```

I realized all of the other tests in MetroArtifactsTest had `multiplatform = false`, but the reportsDestination collision test I added would only make sense having multiplatform set to true. Running `./metrow check` passed locally for me, but I'll see if I can figure this out.

FletchMcKee:
Looked at more of the Gradle plugin tests and I think I see why the ones I added get tripped up with certain Kotlin versions. The KMP gradle tests extend BaseIncrementalCompilationTest because it handles these Kotlin IC issues described in that base class:

```kotlin
/** Minimum Kotlin version that supports incremental compilation for KMP projects. */
private val MULTIPLATFORM_IC_MIN_VERSION = KotlinToolingVersion("2.3.21")

/**
 * Kotlin/JS and Kotlin/Wasm IC trip on top-level declaration generation in these specific Kotlin
 * builds (Metro uses top-level declarations by default for `enableTopLevelFunctionInjection` /
 * `generateContributionHints` / `generateContributionHintsInFir`).
 *
 * See https://youtrack.jetbrains.com/issue/KT-82395 and
 * https://youtrack.jetbrains.com/issue/KT-82989.
 */
private val JS_WASM_IC_TOP_LEVEL_BROKEN_VERSIONS = setOf("2.4.0-Beta1", "2.4.0-dev-2124")
```

This lines up with which versions fail (2.4.0/2.4.20+ pass for Kotlin/JS,  2.4.0-dev-2124 fails). The existing artifact test class doesn’t extend from this base class which makes sense since it would have had collision issues.

So I think if I moved my collision test to a class that uses this base class it should pass, or I could disable those Metro extensions in my added tests. Could also try to make MetroArtifactsTest extend from BaseIncrementalCompilationTest.

gradle-plugin/src/main/kotlin/dev/zacsweers/metro/gradle/MetroGradleSubplugin.kt
ZacSweers, line 1: Let's keep the aggregate report tasks registered/skipped unconditionally, but then only register/wire `MetroArtifactCopyTask` when `reportsDestination.isPresent`? since that task depends on compilation and shouldn't exist on the disabled path

FletchMcKee: That makes sense, if a user runs the task without reportsDestination present they would still see the `onlyIf` message telling them why it was skipped, and their task graph won’t be polluted with all of the copy tasks.

CHANGELOG.md
```diff
@@ -89,6 +90,7 @@ Special thanks to the following contributors for contributing to this release!
 
 - [@BraisGabin](https://github.com/BraisGabin)
 - [@ychescale9](https://github.com/ychescale9)
+- [@FletchMcKee](https://github.com/FletchMcKee)
```
ZacSweers, line 92:
```suggestion
- [@BraisGabin](https://github.com/BraisGabin)
- [@FletchMcKee](https://github.com/FletchMcKee)
- [@ychescale9](https://github.com/ychescale9)
```

PR #2458, opened 2026-06-18
Fix transitive managed containers from instance containers

### LINKED ISSUE #2457: Bug with include BindingContainer defined as `class`
### Description

Metro fails when I contribute a BindingContainer that includes another BindingContainer defined as `class` (instead of `object`).

When running the code below I get this stacktrace:

```
e: file:///Users/brais.gabin/Workspace/tmp/double-scope/src/main/kotlin/FeatureGraph.kt:19:5 [Metro/MissingBinding] Cannot find an @Inject constructor or @Provides-annotated function/property for: com.sample.A

    com.sample.A is injected at
        [com.sample.CommonGraph] com.sample.A.foo(…, <this>)
    com.sample.Foo is requested at
        [com.sample.CommonGraph] com.sample.CommonGraph.foo
```

There are a lot of ways to workaround this:
- Change `A` from `class` to `object` (or even an `abstract class` with `companion object`)
- Don't use `@ContributsTo` and add `B` directly in the `DependencyGraph`
- Don't use `include` and add `A` in `DependencyGraph` or adding `@ContributsTo`

But with this configuration (that I think it is correct) I get that stacktrace.

### Self-contained Reproducer

```kotlin
class Foo

interface CommonGraph {
    val foo: Foo
}

@BindingContainer
class A {
    @Provides
    fun foo(): Foo = Foo()
}

@BindingContainer(includes = [A::class])
class B
```

### Metro environment

```
Project
  path: :
  target: 
  compilation: main
  platform: jvm
  compile task: compileKotlin

Versions
  Metro Gradle plugin: 1.2.1
  Metro compiler artifact: dev.zacsweers.metro:compiler:1.2.1
  Kotlin Gradle plugin: 2.4.0
  Kotlin compiler: 2.4.0
  Gradle: 9.3.0
  Java: 25
  OS: Mac OS X 26.5.1 (aarch64)

Kotlin compiler options
  languageVersion: <default>
  apiVersion: <default>
  freeCompilerArgs:
    -Xcompiler-plugin-order=dev.zacsweers.metro.compiler>androidx.compose.compiler.plugins.kotlin

Metro compiler plugin options
  options:
    enabled = true
    max-ir-errors-count = 20
    debug = false
    generate-assisted-factories = false
    generate-contribution-hints = true
    generate-contribution-hints-in-fir = true
    generate-classes-in-ir = false
    statements-per-init-fun = 25
    enable-graph-sharding = true
    keys-per-graph-shard = 2000
    enable-switching-providers = false
    optional-binding-behavior = DEFAULT
    public-scoped-provider-severity = NONE
    non-public-contribution-severity = NONE
    warn-on-inject-annotation-placement = true
    interop-annotations-named-arg-severity = NONE
    unused-graph-inputs-severity = WARN
    enable-top-level-function-injection = true
    contributes-as-inject = true
    enable-klib-params-check = false
    patch-klib-params = true
    force-enable-fir-in-ide = false
    compiler-version = 
    compiler-version-aliases = 
    enable-function-providers = true
    desugared-provider-severity = WARN
    generate-contribution-providers = true
    enable-circuit-codegen = false
    plugin-order-set = true
    enable-dagger-runtime-interop = false
    enable-kclass-to-class-interop = false
    interop-include-javax-annotations = false
    interop-include-jakarta-annotations = false
    interop-include-dagger-annotations = false
    interop-include-kotlin-inject-annotations = false
    interop-include-anvil-annotations = false
    interop-include-kotlin-inject-anvil-annotations = false
    enable-dagger-anvil-interop = false
    interop-include-guice-annotations = false
    enable-guice-runtime-interop = false
    interop-include-hilt-annotations = false
```

### Previous working version

N/A

### IDE version

_No response_

### Platform

_No response_

### Context

We have really odd patterns from our time with dagger 😅.

BraisGabin:
Resolves #2457


ZacSweers:
Ahh I see the path this was falling through now

PR #2495, opened 2026-06-27
Support parameterless Binds declarations

### LINKED ISSUE #2493: Consider supporting parameter-less `@Binds`
New in Dagger 2.60: https://dagger.dev/dev-guide/#parameterless-binds. Basically seems to be a way to reserve a binding?

maksym-moroz:
Closes #2493.

## Summary

This adds support for Dagger 2.60-style parameterless `@Binds` declarations.

The important semantic distinction is that parameterless `@Binds` is not modeled as a `Foo -> Foo` alias and is not treated as a provider function. Instead, it explicitly claims the returned concrete class's existing `@Inject` constructor binding in the graph so duplicate and override behavior sees it as a declared binding.

## Changes

- Validate parameterless `@Binds` declarations in FIR with Dagger-compatible constraints.
- Add a separate `InjectConstructorBindsCallable` IR model, keeping classic `BindsCallable` source-backed.
- Thread parameterless binds through binding mirrors, graph nodes, inherited graph data, and dynamic container handling.
- Materialize the normal `ConstructorInjected` binding during graph generation and tag it with the `@Binds` declaration for diagnostics.
- Avoid source-parameter body generation for KLIB fake overrides when the declaration is parameterless.
- Update docs/KDoc and add focused compiler tests.

### DIFF
--- a/compiler-tests/src/test/data/box/bindingcontainers/ParameterlessBinds.kt
+++ b/compiler-tests/src/test/data/box/bindingcontainers/ParameterlessBinds.kt
@@ -1,0 +1,20 @@
+@DependencyGraph(bindingContainers = [Bindings::class])
+interface AppGraph {
+  val validFoo: ValidFoo
+}
+
+@BindingContainer
+interface Bindings {
+  @Binds fun bindValidFoo(): ValidFoo
+}
+
+@Inject
+class ValidFoo {
+  val value: String = "constructor"
+}
+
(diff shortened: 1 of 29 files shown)


ZacSweers:
See this repo's contributing guidelines and code of conduct for future contributions, AI generated drive-bys with no acknowledgement like this are not welcome here.

I have a separate branch in process for this anyway.

PR #2535, opened 2026-07-11
Fix JS box test flakiness

kyay10:
The solution is to give each test class its own output directory, so that no races happen while trying to copy files in parallel.

In case you approve workflows to run,`testContributedAccessorsCanBeLazy` keeps failing because the `main` I'm basing this on has it failing, too. Once you fix `main`, update this branch, then workflows will pass.


kyay10:
I forgot that full js box tests don't run on PRs, only smoke ones, so the all-green here is a bit misleading. Still, maybe a good sign, since I removed retries, and it didn't break.

PR #2577, opened 2026-07-22
Android 37 deps

ZacSweers:
All require a newer IJ version


gradle/libs.versions.toml
```diff
@@ -13,15 +13,15 @@
 # limitations under the License.
 
 [versions]
-agp = "9.0.1"
-android-compileSdk = "36"
-androidx-lifecycle = "2.10.0"
+agp = "9.3.0"
```
ZacSweers:
```suggestion
agp = "9.3.1"
```

samples/settings.gradle.kts
```diff
@@ -15,7 +15,7 @@ pluginManagement {
   }
   plugins {
     id("com.gradle.develocity") version "4.5.0"
-    id("com.android.settings") version "9.2.1"
+    id("com.android.settings") version "9.3.0"
```
ZacSweers:
```suggestion
    id("com.android.settings") version "9.3.1"
```

PR #2580, opened 2026-07-22
Run JS tests for future versions of Kotlin

kyay10:
IIUC, klibs produced by version `2.x` [can be read](https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format) by `2.x+3` compiler.
Thus, JS tests can run for versions of Kotlin higher than the build version.
Couldn't figure out the version matrix magic to get CI working on js box tests for these new versions. Running locally with `./ metrow check --all`, everything works just fine!


ZacSweers:
Nice!

PR #2648, opened 2026-08-12
Follow @Origin chains in IrContributionMerger

eygraber:
Excluding or replacing a class that a contributing class declares as its @Origin now removes that contribution's generated binding container. The generated MetroContribution marker's own @Origin (pointing at its contributing class) previously shadowed the user-written one, so origin-based excludes/replaces never matched for root graphs.

Benchmarking showed no regression. 

Fixes https://github.com/ZacSweers/metro/discussions/2644

AI was used to investigate and build a PoC.


ZacSweers:
Thanks!

PR #2677, opened 2026-08-19
enableKClassToClassMapKeyInterop skipped for Provider/Lazy-wrapped maps

grandstaish:
`enableKClassToClassMapKeyInterop` doesn't work when Maps with `Class` keys are wrapped in a `Provider`/`Lazy`. The result is that map lookups don't work at runtime because the map keys are all `KClasses` instances. 

Added a failing case to demonstrate.

```
BoxTestGenerated > Interop > Kclass > testClassMapKeyWrappedInterop() FAILED

java.lang.AssertionError: Expected value to be of type <HelloGreeting (Kotlin reflection is not available)>, actual <null>.
      at kotlin.test.DefaultAsserter.fail(DefaultAsserter.kt:16)
      at kotlin.test.Asserter.assertTrue(Assertions.kt:767)
      at kotlin.test.DefaultAsserter.assertTrue(DefaultAsserter.kt:11)
      at kotlin.test.AssertionsKt__AssertionsKt.assertIsOfType(Assertions.kt:123)
      at kotlin.test.AssertionsKt.assertIsOfType(Unknown Source)
      at ClassMapKeyWrappedInteropKt.box(ClassMapKeyWrappedInterop.kt:25)
```

Note: this did work back in Metro version `1.0.1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ Changelog
 **Unreleased**
 --------------
 
+### Fixes
+
+- **[IR/interop]** Fix `Class`-keyed maps wrapped in providers or lazy values when `KClass`/`Class` interop is enabled.
+
+### Contributors
+
+Special thanks to the following contributors for contributing to this release!
+
+- [@grandstaish](https://github.com/grandstaish)
+
 1.4.2
 -----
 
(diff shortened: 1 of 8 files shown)


ZacSweers:
Thanks!

PR #2679, opened 2026-08-19
Fix AssistedInject runtime crash on Kotlin 2.4.20-RC

grandstaish:
We recently updated to Kotlin 2.4.20-RC and started seeing Metro runtime crashes on a class that had an `@AssistedInject` constructor, but no assisted arguments. 

This bug seems to only reproduce when:
- using the latest Kotlin, e.g. `-Pmetro.testCompilerVersion=2.4.20-RC`. 
- the assisted inject class is in another module

Added a failing case to demonstrate.

Feel free to close this if it's too soon for 2.4.20-RC bugs!

```
BoxTestGenerated > Inject > Assisted > testNoAssistedArgsInAssistedInjectAcrossModules() FAILED

java.lang.ClassCastException: class Example cannot be cast to class Example$MetroFactory (Example and Example$MetroFactory are in unnamed module of loader org.jetbrains.kotlin.codegen.GeneratedClassLoader @706e8586)
      at AppGraph$Impl.getFactory(module_main_NoAssistedArgsInAssistedInjectAcrossModules.kt)
      at Module_main_NoAssistedArgsInAssistedInjectAcrossModulesKt.box(module_main_NoAssistedArgsInAssistedInjectAcrossModules.kt:21)
      at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
      at java.base/java.lang.reflect.Method.invoke(Method.java:565)
      at org.jetbrains.kotlin.test.backend.handlers.JvmBoxRunner.runBoxInCurrentProcess(JvmBoxRunner.kt:179)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Changelog
 **Unreleased**
 --------------
 
+### Fixes
+
+- **[IR]** Fix a runtime crash when an assisted-injected class with no assisted parameters is used across modules with IR class generation.
+
 1.4.2
 -----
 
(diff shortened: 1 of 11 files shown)


ZacSweers:
Thanks!

PR #2721, opened 2026-08-24
IDEA plugin:Fix missing binding with type-use qualifier

kevinguitar:
Noticed a bug in the IDEA plugin, when there are type-use qualified bindings, the plugin doesn't seem connecting the contribution with injection.

CleanShot 2026-08-24 at 16 16 00@2x

CleanShot 2026-08-24 at 16 16 26@2x

The `AutoMixPage` is declared as:
```kotlin
@Qualifier
@Target(AnnotationTarget.TYPE, AnnotationTarget.VALUE_PARAMETER)
annotation class AutoMixPageKey

@ContributesBinding(
    scope = MixEditorScope::class,
    binding = binding<@AutoMixPageKey Page<*>>(),
)
class AutoMixPage : Page<AutoMixPageViewModel>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Changelog
 Special thanks to the following contributors for contributing to this release!
 
 - [@grandstaish](https://github.com/grandstaish)
+- [@kevinguitar](https://github.com/kevinguitar)
 
 1.4.2
 -----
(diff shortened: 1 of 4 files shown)


kevinguitar:
> Could not GET 'https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-serialization-compiler-plugin/2.3.0/kotlin-serialization-compiler-plugin-2.3.0.pom'. Received status code 429 from server: Too Many Requests

Whoops 🙈

PR #2767, opened 2026-09-04
Fix docs for ClockWidgetClass to ClockWidget in context parameter generated code

alexvanyo:
Simple doc fix following #1525


ZacSweers:
Thanks!

PR #2810, opened 2026-09-10
Doc fix: Remove duplicated build input mentions in debugging.md

kevinguitar:


docs/debugging.md
kevinguitar: Kept the newer change that covers the build input mention

kevinguitar, line 17: > This is very much a WIP, feedback is welcome!

I can remove this if this is no longer the case 😄

docs/debugging.md
```diff
@@ -25,9 +25,6 @@ metro {
 }
 ```
 
-!!! warning
-    The Kotlin Gradle Plugin does _not_ include file inputs like `reportsDestination` as build inputs, so you may need to compile with `--rerun` to force recompilation after adding this flag.
```
ZacSweers, line 29: mm we have multiple file input types though, tracing being another. Would rather broaden this then rather than delete

kevinguitar: Yeah, it's in trace destination doc as well. Deleting this as there's a duplicated block right above (the only difference is the KGP mention). I left review comments to point this out

ZacSweers: can we move the KGP mention up then?

kevinguitar: Sure

docs/debugging.md
```diff
@@ -17,17 +17,14 @@ In the future, we could possibly explore including information in IR to synthesi
 Similar to Compose, Metro supports a `reportsDestination` property in its Gradle DSL and can output various graph reports to this destination if specified. This is very much a WIP, feedback is welcome!
 
 !!! warning
-    You should _not_ leave this enabled by default as it can be quite verbose and potentially expensive. This property also does not participate in task inputs, so you may need to recompile with `--rerun` to force recompilation after adding this flag.
+    You should _not_ leave this enabled by default as it can be quite verbose and potentially expensive. The Kotlin Gradle Plugin does _not_ include file inputs like `reportsDestination` as build inputs, so you may need to recompile with `--rerun` to force recompilation after adding this flag.
```
ZacSweers: task inputs, not build inputs

ZacSweers:
```suggestion
    You should _not_ leave this enabled by default as it can be quite verbose and potentially expensive. The Kotlin Gradle Plugin does _not_ include file inputs like `reportsDestination` as task inputs, so you may need to recompile with `--rerun` to force recompilation after adding this flag.
```

PR #2842, opened 2026-09-16
Add cycle benchmarks + cache neighbors

ZacSweers:
Dense graph at 256 nodes: 116 -> 15ms, sealing 117 -> 20ms


Pratikshit26 (commented):
Edit review revoked

PR #2852, opened 2026-09-19
Fix false binding-container warnings for multibinding accessors

vRallev:
Public `@Multibinds` members expose graph accessors. Recognize them as accessors when checking contributed interfaces so Metro no longer recommends a conversion to `@BindingContainer` that would remove their public API.


compiler-tests/src/test/data/diagnostic/aggregation/ContributesToBindingContainer.kt
```diff
@@ -36,22 +36,22 @@ interface <!CONTRIBUTES_TO_COULD_BE_BINDING_CONTAINER!>BindsGetter<!> {
 }
 
 @ContributesTo(AppScope::class)
-interface <!CONTRIBUTES_TO_COULD_BE_BINDING_CONTAINER!>MultibindsFunction<!> {
```
vRallev, line 39: So now I'm not sure anymore. In the first commit I implemented the repro, but in the 2nd commit I discovered this. So it makes me think it was by design? 

Even if it was an actual bug, then the existing test coverage is enough and I can drop my first commit.

ZacSweers: it's a bug, I just missed this in the test golden gen

ZacSweers (approved):

